PR #1, opened 2018-07-25
Typos in SCI-F apps

donkirkby:
Singularity looks like an interesting alternative to Docker.

Here are a couple of typo fixes that I noticed as I was reading the documentation.


jmstover (approved):
Thanks!

jchavezlavalle (approved):
awesome, thanks! :+1:

PR #3, opened 2018-08-13
Patch 1

yoann-dufresne:
The previous installation instructions did not correspond to the current singularity repository.
For example, there is no autogen.sh file.  

I propose to only keep the link to the up to date tuto on github

### DIFF
--- a/quick_start.rst
+++ b/quick_start.rst
@@ -20,7 +20,7 @@ system administrator.
 Installation
 ------------
 
-To install Singularity please follow the instructions on our readthedoc pages here :ref:`install Singularity <installation>` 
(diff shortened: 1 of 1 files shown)


jmstover:
Hi @yoann-dufresne ,

  Thanks, but there's a difference between the versions. The 2.x documentation you're patching against doesn't follow the INSTALL.md instructions in the master branch. That's instructions for building Singularity 3.0 which is completely different.

  There was a recent repository structure change. So, the documentation does need modified to be closer to:

```patch
cd singularity
+ 
+ git fetch --all
+ git tag -l
+ git checkout [tag]
+
+ For example:
+    git checkout 2.5.2   - or -
+    git checkout 2.6.0
+
```

@devprincess 

Thanks,
-J

Issue #5, opened 2018-08-22
Link to defaults.py is incorrect


neiltwist:
Line 90 in build_environment.rst refers to a defaults.py, but that file links to a 404 page:

https://github.com/singularityware/singularity-userdocs/blame/master/build_environment.rst#L90

jchavezlavalle:
@neiltwist thanks for pointing that out, I'm updating that one!

PR #6, opened 2018-08-22
Updating defaults.py url

jchavezlavalle:
This fixes issue #5 : `defaults.py` is not present anymore in latest version of Singularity, up to v2.6.0 the location should be https://github.com/singularityware/singularity/blob/2.6.0/libexec/python/defaults.py

### DIFF
--- a/build_environment.rst
+++ b/build_environment.rst
@@ -87,7 +87,7 @@ Environment Variables
 ---------------------
 
 All environmental variables are parsed by Singularity python helper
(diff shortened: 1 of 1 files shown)


jchavezlavalle:
@jmstover  Issue #5 was pointed out a few hours ago: there was a dead link to `defaults.py` (since it's from master) and I added this one redirecting to its location in `2.6.0`. Is this correct? thanks

jmstover (approved):

PR #7, opened 2018-08-24
Update install

jmstover:
Update more sections of the install instructions


installation.rst
```diff
@@ -106,33 +100,15 @@ releases <https://github.com/singularityware/singularity/releases>`_ page to ``/
...
+The primary development of Singularity now happens on the ``master`` branch.
+Please see the ``INSTALL.md`` file in a copy of the repository.
 
 
 ---------------------
 Remove an old version
 ---------------------
 
-Let’s say that we installed Singularity to ``/usr/local``. To remove it completely,
+Let's say that we installed Singularity to ``/usr/local``. To remove it completely,
```
jchavezlavalle, line 122: Rst is very sensitive with characters, could you rollback this change? other than this I don't see other problems @jmstover

jchavezlavalle (changes requested):

installation.rst
GodloveD: Wait, are these for the 2.6 version of the docs?  Shouldn't we just be telling users to checkout out the 2.6 tag?  These look like instructions for installing 3.0.

installation.rst
```diff
@@ -48,35 +48,29 @@ Install the master branch
 
 The following commands will install the latest version of the `GitHub
```
GodloveD: Let's just change this to say that 
```
The following commands will install version 2.6
```
Or something along those lines.  And then just add the `git checkout 2.6` command.  These are the 2.6 docs after all.

GodloveD (changes requested):

GodloveD:
Also, thanks very much for grabbing this @jmstover

installation.rst
jmstover: The docs are mixed here because of _how_ they are written. They need a thorough go through, and I'm just trying to patch them for now.

jmstover: @devprincess  ... No, I can't revert the character. For me in the RST file, it shows up as `â<80><99>` .... pasting that back in gives that literal string, and not the value you need.

installation.rst
```diff
@@ -112,31 +111,15 @@ releases <https://github.com/singularityware/singularity/releases>`_ page to ``/
 Install the development branch
```
GodloveD, line 111: I think this line needs to be changed right?  There is no such thing as the development branch right?

GodloveD (changes requested):

installation.rst
jchavezlavalle: @GodloveD hey Dave, no there's no dev branch by the moment, all branches correspond to the respective releases by now and master has all the changes up-to-date (so it's like master is a dev branch by the moment)

GodloveD: Oh sorry. I read this in context and it looks good to me.

GodloveD (approved):

PR #19, opened 2018-10-08
added sign, verify, and keys, example page

GodloveD:
Heads up @devprincess!

### DIFF
--- a/signNverify.rst
+++ b/signNverify.rst
@@ -1,0 +1,144 @@
+.. _signNverify:
+
+================================
+signing and verifying containers
+================================
+
+.. _sec:signNverify:
+
+Singularity 3.0 introduces the abilities to create and manage PGP keys and use
+them to sign and verify containers. This provides a trusted method for 
+Singularity users to share a container ensuring a bit-for-bit reproduction of 
+the original container as the author intended it.
+
+-----------------------------------------------
+Verifying containers from the Container Library
+-----------------------------------------------
+
+The ``verify`` command will allow you verify that a container has been signed 
+using a PGP key.  To use this feature with images that you pull from the
+container library, you must first generate an access token to the Sylabs Cloud.
+If you don't already have a valid access token, follow these steps:
+
+  1) Go to : https://cloud.sylabs.io/
+  2) Click "Sign in to Sylabs" and follow the sign in steps.
+  3) Click on your login id (same and updated button as the Sign in one).
+  4) Select "Access Tokens" from the drop down menu.
+  5) Click the "Manage my API tokens" button from the "Account Management" page.
+  6) Click "Create".
+  7) Click "Copy token to Clipboard" from the "New API Token" page.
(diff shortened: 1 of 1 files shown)


signNverify.rst
```diff
@@ -0,0 +1,144 @@
...
+.. code-block:: none
+
+    $ singularity pull library://alpine
+
+    $ singularity verify alpine_latest.sif 
+    Verifying image: alpine_latest.sif
+    Data integrity checked, authentic and signed by:
+    	Sylabs Admin <support@sylabs.io>, KeyID 51BE5020C508C7E9
+
+In this example you can see that Sylabs Admin has signed the container.
```
yhcote: ... and that the locally computed hash matches the one signed at creation.

(Or something that means that)

signNverify.rst
```diff
@@ -0,0 +1,144 @@
...
+
+---------------------------
+Signing your own containers
+---------------------------
+
+Generating and managing PGP keys
+================================
+
+To sign your own containers you first need to generate one or more keys.
```
yhcote, line 53: Doing "singularity sign x.sif" will guide the user go though all that simply by answering questions on the command line if no keys are found. Do you wanna mention that and then say, this is how you do it manually or when you add more keys?

signNverify.rst
```diff
@@ -0,0 +1,144 @@
...
+
+    $ singularity keys list 
+    Public key listing (/home/david/.singularity/sypgp/pgp-public):
+
+    0) U: Dave Godlove (demo) <d@sylabs.io>
+       C: 2018-10-08 15:25:30 -0400 EDT
+       F: 135E426D67D8416DE1D6AC7FFED5BBA38EE0DC4A
+       L: 4096
+       --------
+
```
jchavezlavalle, line 89: @GodloveD I think what's missing is saying F goes for fingerprint and then something like "you should copy this fingerprint to use it as a key at the moment of pushing the container" Not so sure if other fields would be important to mention as well ( U user, C Creation date, F fingerprint as said and L for length of the key)

jchavezlavalle (changes requested):

yhcote (changes requested):
Just a few suggestions.

signNverify.rst
jchavezlavalle: Oh, also forgot on this same thing to add something like.." the `singularity keys list` command will list you the keys you have generated", although evident I think is important to be pointed out to the users, otherwise they could expect just one key at output all the time but in reality they can generate many pairs

signNverify.rst
GodloveD: Nice. Thanks

signNverify.rst
GodloveD: Right on both counts.

yhcote (approved):
Looks good to me!

jchavezlavalle (approved):

PR #20, opened 2018-10-10
Create a security options page

GodloveD:


cclerget (changes requested):
For capabilities you can mention that `capability add/drop`, `--add-caps` and `--drop-caps` support the keyword case insensitive `all` to respectively add/drop all capabilities. Overall looks good to me !

jchavezlavalle (approved):
Just needs the explanation to drop/add all capabilities as suggested from @cclerget and it's good, I will add the note block on the places where notes are mentioned (e.g. "Note that attempting to use the blacklisted...") :+1:

security_options.rst
```diff
@@ -0,0 +1,216 @@
...
+to open raw sockets within Singularity containers, they can revoke the 
+appropriate Linux capability like so:
+
+.. code-block:: none
+
+    $ sudo singularity capability drop --user david CAP_NET_RAW
+
+The ``capabiltiy add`` and ``drop`` subcommands will also accept the case
+insensitive keyword ``all`` to grant or revoke all Linux capabilities to a user
+or group.  Similarly, the ``--add-caps`` option will accept the ``all`` keyword.
```
cclerget, line 63: the ``--add-caps`` and ``--drop-caps`` options

jchavezlavalle: @GodloveD I think only the `--drop-caps` is missing? thank you

GodloveD: Whoops!  I actually added it here:

https://github.com/sylabs/singularity-userdocs/pull/20/files#diff-1a0df1ea1aed450f6c65389ee194f0ffR136

But I previously had the name of the option wrong!  If figured it would be best to talk about the `all` keyword being applicable to the `--drop-caps` command in the section where it is covered.

jchavezlavalle: Looks good then, sorry didn't find it in the same section but you're right, better to add it on its own command section :+1:

GodloveD: Well the command was named improperly, so you cannot be blamed for missing it.  😊

PR #22, opened 2018-10-12
edited build page

GodloveD:


jchavezlavalle (commented):
I think this section needs to be added in the `index.rst` as well, besides that it lgtm @GodloveD

build_a_container.rst
```diff
@@ -0,0 +1,255 @@
...
+
+-------------
+Build options
+-------------
+
+``--builder``
+=============
+
+Singularity 3.0 introduces the option to perform a remote build. The 
+``--builddir`` option allows you to specify a URL to a different build service.
```
ikaneshiro: Should this be ``--builder``? Also, does this need to be used in conjunction with `--remote` like `--detached`?

build_a_container.rst
```diff
@@ -0,0 +1,255 @@
...
+=============
+
+Build a sandbox (chroot directory) instead of the default SIF format.
+
+``--section``
+=============
+
+Instead of running the entire definition file, only run a specific section or 
+sections.  This option accepts a comma delimited string of definition file
+sections.  Acceptable arguments are: ``setup``, ``post``, ``files``,
```
ikaneshiro: Might want to say something along the lines of:
Acceptable arguments include: ``all``, ``none,`` or any combination of valid definition section names. e.g. ``files,environment,post``

build_a_container.rst
```diff
@@ -0,0 +1,255 @@
...
+``--update``
+===========
+
+You can build into the same container multiple times (though the
+results may be unpredictable and it is generally better to delete your
+container and start from scratch).
+
+By default if you build using an existing container name container, the 
+``build`` command prompt you to decide whether or not to overwrite the 
+container. Instead of this behavior you can use the ``--update`` option to build
```
ikaneshiro: We should mention that the update option is only for sandboxes

ikaneshiro (changes requested):

build_a_container.rst
GodloveD: Oh I knew we talked about that but didn't realize that was the final decision.  Thank you.

build_a_container.rst
```diff
@@ -0,0 +1,258 @@
...
+=============
+
+Build a sandbox (chroot directory) instead of the default SIF format.
+
+``--section``
+=============
+
+Instead of running the entire definition file, only run a specific section or 
+sections.  This option accepts a comma delimited string of definition file
+sections.  Acceptable arguments includ ``all``, ``none`` or any combination of
```
ikaneshiro, line 221: small typo includ->include

ikaneshiro (approved):

jchavezlavalle (approved):

build_a_container.rst
jchavezlavalle: I'll fix this in the generation of these docs, thanks @ikaneshiro

PR #24, opened 2018-10-16
Network page

GodloveD:


networking.rst
```diff
@@ -0,0 +1,159 @@
...
+
+======================
+Network virtualization
+======================
+
+.. _sec:networking:
+
+Singularity 3.0 introduces full integration with 
+`cni <https://github.com/containernetworking/cni>`_, and several new features to
+make network virtualisation easy.
```
jchavezlavalle: Change `virtualisation` to `virtualization`

networking.rst
```diff
@@ -0,0 +1,159 @@
...
+
+A few new options have been added to the action commands (``exec``, ``run``, 
+and ``shell``) to facilitate these features, and the ``--net`` option has been
+updated as well.  These options can only be used by root.
+
+``--net``
+=========
+
+Passing the ``--net`` flag will cause the container to join a new network
+network namespace when it initiates.  New in Singularity 3.0, a bridge
```
jchavezlavalle: double word `network` here

networking.rst
```diff
@@ -0,0 +1,159 @@
...
+
+    $ sudo singularity exec --hostname hal-9000 my_container.sif hostname
+    hal-9000
+
+``--network``
+=============
+
+The ``--network`` option can only be invoked in combination with the ``--net``
+flag.  It accepts a comma delimited string of network types. Each  entry will 
+bring up a dedicated interface inside container.
```
jchavezlavalle, line 68: Additional space between `Each` and `entry`

jchavezlavalle (changes requested):

networking.rst
jchavezlavalle: Maybe this section for `--network` flag should be just right after `--net` since they are related ?

jchavezlavalle (changes requested):

networking.rst
GodloveD: I'm not sure about this.  The `--network-args` option is also related in the sense that it doesn't function properly without the `--net` flag.  I basically just promoted the `--net` flag to the top of the list since it was present in 2.x and then alphabetized everything else.  Maybe it's best to just alphabetize them?

jchavezlavalle: yes, good idea :+1: thanks @GodloveD

cclerget (approved):
Looks good to me !

jchavezlavalle (approved):

PR #26, opened 2018-10-18
adding persistent overlay page

GodloveD:
just needed to be updated to use directory instead of ext3 image


jchavezlavalle:
looks good to me :+1: thanks!

PR #27, opened 2018-10-26
added a cgroups page

GodloveD:
@cclerget I've created this page with an example for limiting memory and then checking to see that memory has been limited inside of the running container.  When you have a chance, could you fill in similar examples for the CPU and IO sections?  And would you also mind terribly reviewing the work that I've done? Merci mon ami!

### DIFF
--- a/cgroups.rst
+++ b/cgroups.rst
@@ -20,7 +20,7 @@ apply the settings in the TOML file by using the path as an argument to the
 
 .. code-block:: none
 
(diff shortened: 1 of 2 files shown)


GodloveD:
Thanks very much for the extra examples @cclerget!  This should be RFR @devprincess!

jchavezlavalle:
Just checked and everything looks great, thanks @GodloveD :+1:

PR #28, opened 2018-11-07
bind_paths_and_mounts: formatting error

0xaf1f:
This is a pretty simple change to fix the monospace formatting. I was actually looking through the site looking for a place to add documentation about converting local docker images, but it seems that not everything from the old site has been migrated over to here yet, like http://singularity.lbl.gov/docs-recipes.


jmstover (approved):

jmstover:
Thanks

PR #29, opened 2018-11-13
Add a missing dependency (squashfs-tools)

hasegaw:


jmstover:
@hasegaw Thanks for the note on not having squashfs-tools mentioned.

@devprincess @GodloveD ....  squashfs-tools is a image build dependency, not a Singularity build or run dependency.

Should that be included in this section which is build dependencies? Or mentioned in a different section? The squashfs Kernel module is included in the `kernel` package, so this is only an image build dependency.

jchavezlavalle:
@jmstover @GodloveD Yes I think it can be added here, should I mention below this is an image build dependency only and is not required for singularity build and run commands?

jmstover:
Yeah...  IMO it's good to mention what is an image build dependency, compared to a run, or Singularity compile dependency.

jchavezlavalle (approved):

GodloveD:
I think this is fine too.

GodloveD (approved):

PR #33, opened 2018-11-29
Installation

GodloveD:
Now you know what I've been doing for the last 2 days.  Whew!  

@yhcote it would be really great if you could go over the `mconfig` bits and keep me honest!

Thank you!

### DIFF
--- a/installation.rst
+++ b/installation.rst
@@ -218,7 +218,7 @@ build and install the RPM.
 
 .. code-block:: none
 
-    $ export VERSION=3.0.1 # adjust this as necessary && \
+    $ export VERSION=3.0.1 && # adjust this as necessary \
         wget https://github.com/sylabs/singularity/releases/download/v${VERSION}/singularity-${VERSION}.tar.gz && \
         rpmbuild -tb singularity-${VERSION}.tar.gz && \
         sudo rpm -ivh ~/rpmbuild/RPMS/x86_64/singularity-$VERSION-1.el7.x86_64.rpm && \
@@ -245,45 +245,33 @@ of the default ``/var``) you can do the following:
 Remove an old version
 =====================
 
-The ``make install`` command lists the files that are installed with 
-Singularity. For instance, by default Singularity 3.0.1 installs the following 
-files. They must all be removed in order to completely remove Singularity. 
+When you run ``sudo make install``, the command lists files as they are 
+installed. They must all be removed in order to completely remove Singularity.
+
+For example, in a standard installation of Singularity 3.0.1 (when building from
+source) you must remove all of these files and directories to completely remove
+Singularity. 
+
+Obviously, this list of files may differ depending on how you install 
+Singularity or with newer versions of Singularity released following the writing
+of this document. 
 
 .. code-block:: none
 
-    $ sudo rm -rf 
+    $ sudo rm -rf \
+        /usr/local/libexec/singularity \
+        /usr/local/var/singularity \
+        /usr/local/etc/singularity \
         /usr/local/bin/singularity \
-        /usr/local/libexec/singularity/bin/starter \
-        /usr/local/libexec/singularity/bin/starter-suid \
-        /usr/local/var/singularity/mnt/session \
-        /usr/local/etc/singularity/singularity.conf \
-        /usr/local/bin/run-singularity \
-        /usr/local/etc/singularity/capability.json \
-        /usr/local/etc/singularity/ecl.toml \
-        /usr/local/etc/bash_completion.d/singularity \
-        /usr/local/etc/singularity/actions \
-        /usr/local/libexec/singularity/cni/bandwidth \
-        /usr/local/libexec/singularity/cni/bridge \
-        /usr/local/libexec/singularity/cni/dhcp \
-        /usr/local/libexec/singularity/cni/flannel \
-        /usr/local/libexec/singularity/cni/host-device \
-        /usr/local/libexec/singularity/cni/host-local \
-        /usr/local/libexec/singularity/cni/ipvlan \
-        /usr/local/libexec/singularity/cni/macvlan \
-        /usr/local/libexec/singularity/cni/portmap \
-        /usr/local/libexec/singularity/cni/ptp \
-        /usr/local/libexec/singularity/cni/static \
-        /usr/local/libexec/singularity/cni/tuning \
-        /usr/local/libexec/singularity/cni/vlan \
-        /usr/local/etc/singularity/seccomp-profiles/default.json \
-        /usr/local/etc/singularity/nvliblist.conf \
-        /usr/local/etc/singularity/cgroups/cgroups.toml
+        /usr/local/bin/run-singularity \ 
+        /etc/bash_completion.d/singularity
 
 If you anticipate needing to remove Singularity, it might be easier to install 
 it in a custom directory using the ``--prefix`` option to ``mconfig``.  In that 
 case Singularity can be uninstalled simply by deleting the parent directory. Or 
 it may be useful to install Singularity :ref:`using a package manager 
-<install-rpm>` so that it can be uninstalled with ease.   
+<install-rpm>` so that it can be updated and/or uninstalled with ease in the
+future.   
 
 ====================================
 Distribution packages of Singularity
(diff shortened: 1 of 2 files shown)


installation.rst
```diff
@@ -0,0 +1,562 @@
...
+    <https://github.com/sylabs/singularity/releases>`_.  It 
+    should be named `singularity-<version>.tar.gz`.
+
+After installing the :ref:`dependencies <install-dependencies>` and installing 
+:ref:`Go <install-go>` as detailed above, you are ready download the tarball and 
+build and install the RPM.
+
+.. code-block:: none
+
+    $ export VERSION=3.0.1 # adjust this as necessary && \
```
jmstover: This won't work for copy paste. Change to:
```
$ export VERSION=3.0.1 && # adjust this as necessary \
```

installation.rst
```diff
@@ -0,0 +1,562 @@
...
+=====================
+
+The ``make install`` command lists the files that are installed with 
+Singularity. For instance, by default Singularity 3.0.1 installs the following 
+files. They must all be removed in order to completely remove Singularity. 
+
+.. code-block:: none
+
+    $ sudo rm -rf 
+        /usr/local/bin/singularity \
```
jmstover, line 266: You don't have the `run-singularity` script

jmstover (changes requested):

installation.rst
GodloveD: In my hands it works just fine. (I double checked b/c I also had doubts.) But maybe it doesn't work on old crusty SlackWear or whatever so I'm happy to change. 

😂

installation.rst
GodloveD: So that means that it's not being listed by `sudo make install`.  meh.  Wonder what else is being silently installed.  😞

GodloveD: Do you think I should address this by just instructing users to delete all of the parent directories that are created during installation?

jmstover: I just saw that it's there... the list is out of order. If you just copy pasted that makes sense though. ;)

You can remove all of:
  `/usr/local/libexec/singularity`
  `/usr/local/etc/singularity`

But having a removal of `/usr/local/bin` ... not so much. ;P 

See this thread:
https://groups.google.com/a/lbl.gov/forum/#!msg/singularity/B28go-RUpXU/Bztw1dimAgAJ

Add in the file: `$prefix/etc/bash_completion.d/singularity`

And a possibility of files being under:  `$prefix/share/man/man1`  (or if they set the `--mandir` option... `$(MANDIR)/man1`)

installation.rst
jmstover: If that works ... they changed behavior of comments. I don't like that...   I mean, you're commenting out the `&&` ... 

For me, I can't run it with the `&& \` behind the `#`  ... but as long as the `&&` is before the `#` it works. I mean ... That should be expected behavior. :/

Does it do any thing stupid if you have a script with a comment like:
```
# this && that here && \
test -f $0 \
echo $?
```

installation.rst
```diff
@@ -0,0 +1,562 @@
...
+it may be useful to install Singularity :ref:`using a package manager 
+<install-rpm>` so that it can be uninstalled with ease.   
+
+====================================
+Distribution packages of Singularity
+====================================
+
+.. note::
+
+    Packaged versions of Singularity in Linux distrubution repos are maintained
```
jchavezlavalle: There's a typo in "distribution" on line 294 here

installation.rst
```diff
@@ -0,0 +1,562 @@
...
+    above.
+
+-----------------------------------------------
+Install the Debian/Ubuntu package using ``apt``
+-----------------------------------------------
+
+Singularity is available on Debian and derivative distributions starting with 
+Debian stretch and the Ubuntu 16.10 releases. The package is called 
+``singularity-container``.  For more recent releases of singularity and 
+backports for older Debian and Ubuntu releases, it is recommend that you use the
```
jchavezlavalle: "...and Ubuntu releases, it is **recommended**..."

installation.rst
```diff
@@ -0,0 +1,562 @@
...
+If you need a backport build of the recent release of Singularity on those or 
+older releases of Debian and Ubuntu, you can `see all the various builds and 
+other information here 
+<http://neuro.debian.net/pkgs/singularity-container.html>`_.
+
+---------------------------------------------
+Install the CentOS/RHEL package using ``yum``
+---------------------------------------------
+
+The epel (Extra Packages for Enterprise Linux) repos contains Singularity. The
```
jchavezlavalle: The epel repos **contain** Singularity

installation.rst
```diff
@@ -0,0 +1,562 @@
...
+
+This document, and the accompanying administrator documentation provides answers
+to many common questions.
+
+If you need to request an installation you may decide to draft a message similar 
+to this:
+
+.. code-block:: none
+
+    Dear shared resource admininstrator,
```
jchavezlavalle: Dear shared resource **administrator**

jchavezlavalle (changes requested):
some typos to check here @GodloveD, thanks! :+1:

installation.rst
GodloveD: I don't think the `make man` command is even working in the 3.0.1 version of Singularity so I'm going to omit telling folks to remove man pages for now.

GodloveD: @jmstover plz resolve this convo if you are happy.

jmstover: It works ... if you build the RPM you have the man pages. 

If you change `_prefix` it breaks, but there is a fix in for that.

jmstover: The bash completion will be at:  `/usr/local/etc/bash_completion.d/singularity` . It gets installed to `$(SYSCONFDIR)/bash_completion.d` ... and a `--prefix=/usr/local` makes that value:  `/usr/local/etc`

GodloveD: Well, this procedure is not for an RPM installation right?  Because then you would just use `rpm -e`

The only reason you would be manually deleting files is if you ran `make install` yourself.  And like I said, `make man` is busted right now.  

Also, as far as the bash completion file, I'm explicitly telling users to copy it to `/etc/bash_completion.d/singularity` so that it will work.

https://github.com/sylabs/singularity-userdocs/blob/df148635d5cf1ce6c3c66d0e7b59b5b04c8027d6/installation.rst#copy-bash-completion-file

Maybe that is why bash completion doesn't work by default with an rpm installation.

jmstover: That's still going to leave the one that's installed under:  `/usr/local/etc/bash_completion.d` though...

That's where `make install` would put a copy in this case

GodloveD: Jeez.  I didn't even know that `make install` put a copy there.  Why does it do so?  It's basically useless there.  It's not like it will make bash completion work in that directory.

jmstover: Just not automatically ... if you source it, then it still works fine.

And it installs it to:  `$(SYSCONFDIR)/bash_completion.d/`

So, for the RPM ... that would be `/etc` by default. In this case it's `$prefix/etc`.

GodloveD: Well, the `singularity` binary would also work if we put it in `/usr/local/bin/singularity/singularity` or some other silly path if we just told users to adjust their `$PATH` but that is obviously quite ridiculous. 

I'm going to change the instructions that I provide so that I'm advising users to source `/usr/local/etc/bash_completion.d/singularity` and to put that change in their `.bashrc`.  But I'm also going to open an issue and advocate that we start doing something more sane with the bash completion file.  This is pretty silly.

jmstover: No... It comes down to the value of `SYSCONFDIR`. If you want it to be system wide, you need to specify a system wide value for that. But that same location also controls where the `singularity.conf` ends up. So, if you want system wide to work out of the box (assuming you have a profile script to source all the /etc/bash_completion.d/ files), do something like:

```
./mconfig --prefix=/usr/local --sysconfdir=/etc
```

That's going to place:  `/etc/bash_completion.d/singularity` ... but it will also place everything under `/etc/singularity`

GodloveD: OK.  Check my latest commit and see if it works better with the current state of things.

jmstover: :+1:

jmstover (approved):

kmanalo:
@GodloveD I was actually looking for this content, so thanks for your great timing

jchavezlavalle (approved):
lgtm! @GodloveD

GodloveD:
Thank you!

PR #35, opened 2018-12-08
Add definition file instructions 

GodloveD:
Another big ole nasty review!  😜

### DIFF
--- a/appendix.rst
+++ b/appendix.rst
@@ -204,7 +204,7 @@ container can be in either image or directory format.
 Overview
 ========
 
-You can use an existing container image as your “base,” and then add 
+You can use an existing container image as your “base”, and then add 
 customization. This allows you to build multiple images from the same starting 
 point. For example, you may want to build several containers with the same 
 custom python installation, the same custom compiler toolchain, or the same base 
--- a/appendix.rst
+++ b/appendix.rst
@@ -277,9 +277,9 @@ use. It is only required if you have specified a %{OSVERSION} variable in the
 
     MirrorURL: http://mirror.centos.org/centos-%{OSVERSION}/%{OSVERSION}/os/$basearch/
 
-The MirrorURL keyword is mandatory. It specifies the URL to use as a mirror to 
+The MirrorURL keyword is mandatory. It specifies the URI to use as a mirror to 
 download the OS. If you define the ``OSVersion`` keyword, than you can use it in 
-the URL as in the example above.
+the URI as in the example above.
 
 .. code-block:: none
 
@@ -354,7 +354,7 @@ to use. For Ubuntu you can use code words like ``trusty`` (14.04), ``xenial``
 
      MirrorURL:  http://us.archive.ubuntu.com/ubuntu/
 
-The MirrorURL keyword is mandatory. It specifies a URL to use as a mirror when 
+The MirrorURL keyword is mandatory. It specifies a URI to use as a mirror when 
 downloading the OS.
 
 .. code-block:: none
@@ -455,7 +455,7 @@ use.
 
     MirrorURL: https://www.busybox.net/downloads/binaries/1.26.1-defconfig-multiarch/busybox-x86_64
 
-The MirrorURL keyword is mandatory. It specifies a URL to use as a mirror when 
+The MirrorURL keyword is mandatory. It specifies a URI to use as a mirror when 
 downloading the OS.
 
 Notes
--- a/appendix.rst
+++ b/appendix.rst
@@ -1,0 +1,508 @@
+========
+Appendix
+========
+
+.. _build-library-module:
+
+---------------------------
+``library`` bootstrap agent
+---------------------------
+
+.. _sec:build-library-module:
+
+Overview
+========
+
+You can use an existing container on the Container Library your “base,” and then 
+add customization. This allows you to build multiple images from the same 
+starting point. For example, you may want to build several containers with the 
+same custom python installation, the same custom compiler toolchain, or the same 
+base MPI installation. Instead of building these from scratch each time, you 
+could create a base container on the Container Library and then build new 
+containers from that existing base container adding customizations in ``%post``, 
+``%environment``, ``%runscript``, etc.
+
+Keywords
+========
+
+.. code-block:: none
+
+    Bootstrap: library
+
+The Bootstrap keyword is always mandatory. It describes the bootstrap module to 
+use.
+
+.. code-block:: none
+
+    From: <entity>/<collection>/<container>:<tag>
+
+The From keyword is mandatory. It specifies the container to use as a base. 
+``entity`` is optional and defaults to ``library``. ``collection`` is 
+optional and defaults to ``default``. This is the correct namespace to use for 
+some official containers (``alpine`` for example). ``tag`` is also optional and 
+will default to ``latest``.
+
+.. code-block:: none
+
+    Library: http://custom/library
+
+The Library keyword is optional. It will default to `` https://library.sylabs.io``.
+
+
+.. _build-docker-module:
+
+--------------------------
+``docker`` bootstrap agent
+--------------------------
+
+.. _sec:build-docker-module:
+
+Overview
+========
+
+Docker images are comprised of layers that are assembled at runtime to create an 
+image. You can use Docker layers to create a base image, and then add your own 
+custom software. For example, you might use Docker’s Ubuntu image layers to 
+create an Ubuntu Singularity container. You could do the same with CentOS, 
+Debian, Arch, Suse, Alpine, BusyBox, etc.
+
+Or maybe you want a container that already has software installed. For instance, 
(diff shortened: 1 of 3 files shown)


appendix.rst
```diff
@@ -0,0 +1,508 @@
...
+``entity`` is optional and defaults to ``library``. ``collection`` is 
+optional and defaults to ``default``. This is the correct namespace to use for 
+some official containers (``alpine`` for example). ``tag`` is also optional and 
+will default to ``latest``.
+
+.. code-block:: none
+
+    Library: http://custom/library
+
+The Library keyword is optional. It will default to `` https://library.sylabs.io``.
```
jchavezlavalle: Shouldn't this url be ``https://cloud.sylabs.io/library`` ?

jchavezlavalle (changes requested):

appendix.rst
```diff
@@ -0,0 +1,508 @@
...
+---------------------------
+``library`` bootstrap agent
+---------------------------
+
+.. _sec:build-library-module:
+
+Overview
+========
+
+You can use an existing container on the Container Library your “base,” and then
```
jmstover:
```
as your "base,"
```

appendix.rst
```diff
@@ -0,0 +1,508 @@
...
+.. code-block:: none
+
+    Bootstrap: library
+
+The Bootstrap keyword is always mandatory. It describes the bootstrap module to 
+use.
+
+.. code-block:: none
+
+    From: <entity>/<collection>/<container>:<tag>
```
jmstover, line 37: Is this still the current layout? I know it's being changed....   @tri-adam

jmstover (changes requested):

definition_files.rst
```diff
@@ -0,0 +1,590 @@
...
+.. code-block:: none
+
+    $ singularity instance start my_container.sif instance1
+    INFO:    instance started successfully
+
+    $ lsof | grep LISTEN
+    nc        19061               vagrant    3u     IPv4             107409      0t0        TCP *:12345 (LISTEN)
+
+    $ singularity instance stop instance1
+    Stopping instance1 instance of /home/vagrant/my_container.sif (PID=19035)
```
jchavezlavalle, line 373: I guess it should be  ``(PID=19061)`` ?

definition_files.rst
```diff
@@ -0,0 +1,590 @@
...
+that software works as expected on custom hardware. If you want to build a 
+container without running the ``%test`` section (for example, if the build
+system does not have the same hardware that will be used on the production 
+system), you can do so with the ``--notest`` build option:
+
+.. code-block:: none
+
+    $ sudo singularity build --notest my_container.sif my_container.def
+
+Running the test command on a container built with this def file yeilds the
```
jchavezlavalle: "..with this def file **yields** the following:"

definition_files.rst
```diff
@@ -0,0 +1,590 @@
...
+.. code-block:: none
+
+    $ singularity exec --app foo my_container.sif env | grep SOFTWARE
+    SOFTWARE=foo
+
+    $ singularity exec --app bar my_container.sif env | grep SOFTWARE
+    SOFTWARE=bar
+
+--------------------------------
+Best Practices for Build Recipes
```
jchavezlavalle, line 567: Best Practices for **Building** Recipes?

jchavezlavalle (changes requested):

definition_files.rst
jmstover: I think "Build Recipes" works in this context.

definition_files.rst
jmstover: No ... 19035 should be the process of the parent instance process, where 19061 is the PID of a process running from within the instance.

definition_files.rst
```diff
@@ -0,0 +1,590 @@
...
+
+Commands in the ``%setup`` section are executed on the host system outside of 
+the container after the base OS has been installed. You can reference the 
+container file system with the ``$SINGULARITY_ROOTFS`` environment variable in 
+the ``%setup`` section. 
+
+.. note::
+
+    Be careful with the ``%setup`` section! This scriptlet is executed outside
+    of the container on the host system itself, and is executed will elevated
```
ikaneshiro: with?

definition_files.rst
```diff
@@ -0,0 +1,590 @@
...
+build. Because of the potential danger involved in running the ``%setup`` 
+scriptlet with elevated priviledges on the host system during the build, it's 
+use is generally discouraged.
+
+%files
+======
+
+The ``%files`` section allows you to copy files from your host system into the 
+container with greater safety than using the ``%setup`` section. Each line is a 
+pair of ``<source>`` and ``<destination>``, where the source is a path on your
```
ikaneshiro: How about:
Each line is a ``<source>`` and ``<destination>`` pair, ...

definition_files.rst
```diff
@@ -0,0 +1,590 @@
...
+In this runscript, the time that the container was created is echoed via the 
+``$NOW`` variable (set in the ``%post`` section above). The options passed to 
+the container at runtime are printed as a single string (``$*``) and then they 
+are passed to echo via a quoted array (``$@``) which ensures that all of the 
+arguments are properly parsed by the executed command. The ``exec`` preceeding 
+the final ``echo`` command replaces the current entry in the process table 
+(which originally was the call to Singularity). Thus the runscript shell process 
+ceases to exist, and only the process running within the container remains.
+
+Running the container built using this def file will yeild the following:
```
ikaneshiro: yield

appendix.rst
```diff
@@ -0,0 +1,508 @@
...
+.. _build-yum:
+
+-----------------------
+``yum`` bootstrap agent
+-----------------------
+
+.. _sec:build-yum:
+
+This module allows you to build a Red Hat/CentOS/Scientific Linux style 
+container from a mirror URI.
```
ikaneshiro, line 250: We reference uri's throughout this doc when talking about things related to `MirrorURL`. Should we change these to `URL` and leave `URI` for things like `docker`, `shub` and `library`?

ikaneshiro (changes requested):

appendix.rst
GodloveD: It's the layout as of now.  We will need to change when it changes (and matching changes have been made and merged and released in Singularity.)

appendix.rst
GodloveD: I'm actually not sure when to say URI and URL.  It's my understanding that if one is in doubt URI is a superset and should be prefered.  Thoughts?

ikaneshiro: I'm pretty sure this counts as a `URL` since it requires `http(s)://`. But you're correct that `URI` is a superset of `URL`, so I think its fine to say that. I'm just suggesting to keep it consistent throughout the page one way or the other. Like if you look at line 280, that section uses `URL` instead of `URI`.

ikaneshiro (approved):

appendix.rst
```diff
@@ -0,0 +1,509 @@
...
+.. _sec:build-localimage:
+
+This module allows you to build a container from an existing Singularity 
+container on your host system. The name is somewhat misleading because your 
+container can be in either image or directory format.
+
+Overview
+========
+
+You can use an existing container image as your “base,” and then add
```
jchavezlavalle: `as your “base”,`

jchavezlavalle (changes requested):

definition_files.rst
```diff
@@ -0,0 +1,590 @@
...
+--------
+
+The main content of the bootstrap file is broken into sections. Different 
+sections add different content or execute commands at different times during the 
+build process. Note that if any command fails, the build process will halt.
+
+Here is an example definition file that uses every available section. We will 
+discuss each section in turn. It is not necessary to include every section (or 
+any sections at all) within a def file. Furthermore, the order of the sections
+in the def file is uninportant and multiple sections of the same name can be
```
jchavezlavalle: in the def file is **unimportant** and multiple...

definition_files.rst
```diff
@@ -0,0 +1,590 @@
...
+
+Here, ``file1`` is created at the root of the file system **on the host**.  
+We'll use ``file1`` to demonstrate the usage of the ``%files`` section below. 
+The ``file2`` is created at the root of the file system **within the 
+container**. 
+
+In later versions of Singularity the ``%files`` section is provided as a safer
+alternative to copying files from the host system into the container during the 
+build. Because of the potential danger involved in running the ``%setup`` 
+scriptlet with elevated priviledges on the host system during the build, it's
```
jchavezlavalle: with elevated **privileges** on the host system during the build, **its** use is generally discouraged

jchavezlavalle (changes requested):

appendix.rst
GodloveD: all of your "base," are belong to us

jmstover (approved):

jchavezlavalle (approved):

Issue #47, opened 2018-12-19
merge 3.0 branch into master and remove 3.0


GodloveD:
I think this is the way to go moving forward to make it easier for users to contribute.

Issue #48, opened 2018-12-19
remove _build directory from repo


GodloveD:
And add `_build` to the `.gitignore` file.  This would make it much easier to work with the repo.

jchavezlavalle:
This has already been solved :+1:

PR #49, opened 2018-12-20
[ Issues #47 and #48 ] Updated master branch to latest 3.0 and ignored build files that are not necessary

jchavezlavalle:
- Fixes: #47 and #48 

Please notice that `_build` folder still exists but it has only 1 single folder needed: `_build/html/static/js` which contains only additional js code which can be updated later for our site to add additional features.

For example our footer: 
image

Has been coded with -> `_build/html/static/js/footer.js`


jchavezlavalle:
@GodloveD This is RFR, contains @WestleyK  's changes on Contributing section and missing .github files which have been merged from 3.0 branch

WestleyK (approved):
Looks good to me! Imo...

GodloveD (approved):
lgtm

hall:
@devprincess, instead of modifying the build output directory, you can add a [method](https://docs.readthedocs.io/en/latest/guides/adding-custom-css.html) to `conf.py` that will include static content:
```py
def setup(app):
    app.add_stylesheet('custom.css')
    app.add_javascript('custom.js')
```
Then just add those files (`custom.css` and `custom.js`) under the [`html_static_path`](https://github.com/sylabs/singularity-userdocs/blob/master/conf.py#L145) and it should work as expected.

GodloveD:
Oh that sounds like a great idea @hall!  What do you think @devprincess?

PR #55, opened 2018-12-30
Add EPUB documentation builder

dmitryilyin:
## Add EPUB format builder configuration

While PDF is awesome is many cases it does not work really well on the small screens
of mobile devices. Let's add EPUB format too.

Also, could you please add EPUB files to https://www.sylabs.io/docs/
and please rename PDF files to include version numbers i.e. user-guide-3.0.pdf

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -273,3 +273,46 @@ texinfo_documents = [
 
 # If true, do not generate a @detailmenu in the "Top" node's menu.
 #texinfo_no_detailmenu = False
+
+# -- Options for Epub output -------------------------------------------------
+
+# Bibliographic Dublin Core info.
+
+# The title of the document. It defaults to the html_title option
+# but can be set independently for epub creation.
+epub_title = 'Singularity ' + version + ' User Documentation'
+
+# The author of the document. This is put in the Dublin Core metadata.
+# It defaults to the author option.
+epub_author = project
+
+# The publisher of the document. This is put in the Dublin Core metadata.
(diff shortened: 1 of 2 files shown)


conf.py
```diff
@@ -273,3 +273,46 @@
...
 # If true, do not generate a @detailmenu in the "Top" node's menu.
 #texinfo_no_detailmenu = False
+
+# -- Options for Epub output -------------------------------------------------
+
+# Bibliographic Dublin Core info.
+
+# The title of the document. It defaults to the html_title option
+# but can be set independently for epub creation.
+epub_title = 'Singularity User Docs ' + version
```
jmstover: Can you change this to:

```
epub_title = 'Singularity User Documentation ' + version
```

I'm also somewhat thinking maybe:
```
'Singularity ' + version + ' User Documentation'
```
But, that's a semantics thing on how it gets viewed. The version is the Singularity version, not a documentation versioning... so...  @devprincess @GodloveD

jmstover (dismissed):

conf.py
dmitryilyin: Done

jmstover (approved):
LGTM... @GodloveD or @devprincess do with it as you will.

GodloveD (approved):
Also lgtm.  @devprincess

GodloveD:
Thanks very much for these contributions @dmitryilyin !

jchavezlavalle:
lgtm as well, thanks @dmitryilyin !

PR #91, opened 2019-02-07
First draft on Build environment section -> 3.0

### LINKED ISSUE #85: build environment page

jchavezlavalle:

First draft on build environment section for 3.0 

- Fixes #85


build_env.rst
```diff
@@ -0,0 +1,150 @@
...
+
+To make download of layers for build and :ref:`pull <pull-command>` faster and less redundant, we
+use a caching strategy. By default, the Singularity software will create
+a set of folders in your ``$HOME`` directory for docker layers, Cloud library images and metadata, respectively:
+
+.. code-block:: none
+
+    $HOME/.singularity/oci
+    $HOME/.singularity/oci-tmp
+
```
WestleyK: Again, I would add the `library` here:

```suggestion 
    $HOME/.singularity/cache/library
```

WestleyK (changes requested):
One small change, otherwise it looks good!

build_env.rst
```diff
@@ -0,0 +1,150 @@
+.. _build-environment:
+
+=================
+Build Environment
+=================
+
+.. _sec:buildenv:
```
GodloveD, line 7: please add on overview section.  😸

build_env.rst
GodloveD: I was just about to comment that this discussion seems to ommit the library cache.  I see that @WestleyK has beaten me to it. 

Here's what I'm seeing in the cache after building alpine images from docker hub, library, and shub:

```
vagrant@vagrant:~$ tree .singularity/cache/
.singularity/cache/
├── library
│   └── sha256.69ce2a3dcc6d3e559e20ced0df251046ee6ecff390a945d856fe0dcb3bcb3ce8
│       └── alpine_latest.sif
├── oci
│   ├── blobs
│   │   └── sha256
│   │       ├── 57c41aa0b96eb7fcf351321550126306e1da1bc1b1c1c847f4976fd201d95ba1
│   │       ├── 650605be53547aeeb66575b82b865ac86f8f4e3e5026ee7f0305426dcaae9ea6
│   │       └── 6c40cc604d8e4c121adcb6b0bfe8bb038815c350980090e74aa5a6423f8f82c0
│   ├── index.json
│   └── oci-layout
└── oci-tmp
    └── b3dbf31b77fd99d9c08f780ce6f5282aba076d70a513a8be859d8d3a4d0c92b8
        └── alpine_latest.sif
```

build_env.rst
```diff
@@ -0,0 +1,150 @@
...
+a set of folders in your ``$HOME`` directory for docker layers, Cloud library images and metadata, respectively:
+
+.. code-block:: none
+
+    $HOME/.singularity/oci
+    $HOME/.singularity/oci-tmp
+
+
+If you want replace the full path where you want to cache, set ``SINGULARITY_CACHEDIR`` to the desired path.
+Remember that when you run commands as sudo this will use root’s home at ``/root`` and not your user’s home.
```
GodloveD, line 33: Maybe it is also useful to add a note here about using the `-E` option when running as root so that `SINGULARITY_CACHEDIR` will be passed along and respected.

build_env.rst
```diff
@@ -0,0 +1,150 @@
...
+-----------------
+
+ .. _sec:temporaryfolders:
+
+Singularity also uses some temporary directories to build the squashfs filesystem,
+so this temp space needs to be large enough to hold the entire resulting Singularity image.
+By default this happens in ``/tmp`` but can be overridden by setting ``SINGULARITY_TMPDIR`` to the full
+path where you want the squashfs temp files to be stored. Since images
+are typically built as root, be sure to set this variable in root’s
+environment.
```
GodloveD: Or use the `-E` option with `sudo`.  (I think.  Haven't actually tested that this works, but I see no reason it would not.)

build_env.rst
```diff
@@ -0,0 +1,150 @@
...
+:ref:`pull docs <pull-command>`. You have the similar ability to set it to be something
+different, or to customize the name of the pulled image.
+
+---------------------
+Environment Variables
+---------------------
+
+All environmental variables are parsed by Singularity python helper
+functions, and specifically the file `defaults.py <https://github.com/sylabs/singularity/blob/2.6.0/libexec/python/defaults.py>`_ is a gateway
+between variables defined at runtime, and pre-defined defaults. By way
```
GodloveD: This is no longer true in 3.0.  Come to think of it, I don't think it was ever true.  Maybe build-specific env vars were parsed in python, but many env vars were simply parsed using bash.  Now they are all parsed using Cobra in Go.  Defaults are set in the cli code. 

Now the order of operations is as follows.  

1.  If a flag is represented by both a CLI option and a env var, and both are set, the CLI option takes precedence.  This is true for all env vars except for `SINGULARITY_BIND`/`SINGULARITY_BINDPATH` which is _combined_ with the `--bind` option, argument pair if both are present.
2. env vars overwrite default values in the CLI code
3. Any defaults in the CLI code are applied.

build_env.rst
```diff
@@ -0,0 +1,150 @@
...
+
+#. Variables that should not be displayed in debug logger are set with ``silent=True``,
+   and are only reported to be defined.
+
+For boolean variables, the following are acceptable for True, with any
+kind of capitalization or not:
+
+.. code-block:: none
+
+    ("yes", "true", "t", "1","y")
```
GodloveD: I don't think this is true anymore.  Don't know how we are handling bools.  I _think_ it should be that non-empty indicates `true`.  So if you set `export SINGULARITY_SOMEBOOL=false` it should be set as true because it is non-empty.  Should be tested I think.

build_env.rst
```diff
@@ -0,0 +1,150 @@
...
+-----
+
+The location and usage of the cache is also determined by environment
+variables.
+
+**SINGULARITY_CACHEDIR** Is the base folder for caching layers and
+singularity hub images. If not defined, it uses default of ``$HOME/.singularity``. If
+defined, the defined location is used instead.
+
+**SINGULARITY_PULLFOLDER** While this isn’t relevant for build, since
```
GodloveD, line 83: Please test everything to make sure it is working as you write docs.  See: https://github.com/sylabs/singularity/issues/2788

build_env.rst
```diff
@@ -0,0 +1,150 @@
...
+Defaults
+========
+
+The following variables have defaults that can be customized by you via
+environment variables at runtime.
+
+Docker
+------
+
+**DOCKER_API_BASE** Set as ``index.docker.io``, which is the name of the registry. In
```
GodloveD: @devprincess can you please test and verify whether the `DOCKER*` vars actually do anything with respect to 3.0?  I'm guessing they probably do not.

build_env.rst
```diff
@@ -0,0 +1,150 @@
...
+choose an architecture (this variable) and an os (above). The default
+is ``amd64``, and other common ones include ``arm``, ``arm64``, ``ppc64le``, ``386``, and ``s390x``.
+**NAMESPACE** Is the default namespace, ``library``.
+
+**RUNSCRIPT_COMMAND** Is not obtained from the environment, but is a
+hard coded default (“/bin/bash”). This is the fallback command used in
+the case that the docker image does not have a CMD or ENTRYPOINT.
+**TAG** Is the default tag, ``latest``.
+
+**SINGULARITY_NOHTTPS** This is relevant if you want to use a
```
GodloveD, line 112: There are a few other variables that are very important here right?  
```
SINGULARITY_DOCKER_LOGIN
SINGULARITY_DOCKER_USERNAME
SINGULARITY_DOCKER_PASSWORD
```
Also there are some variables that are important for building from library:
```
SINGULARITY_BUILDER
SINGULARITY_LIBRARY
SINGULARITY_REMOTE
```
And maybe there are some other vars I'm forgetting?

GodloveD (changes requested):
Heya @devprincess.  Sorry it took me so long to get around to reviewing this page!!  😩 

Looks like it still needs some love.  In particular, I'm afraid that it has some info in it that is no longer current as of 3.0.

WestleyK (approved):
Looks Good to me!!! :+1:

PR #96, opened 2019-02-13
Add auto-generated CLI docs

### LINKED ISSUE #73: Add CLI reference
The old docs had a very useful CLI reference: https://singularity.lbl.gov/docs-usage. However, this is not present in the new user guide: https://www.sylabs.io/guides/3.0/user-guide/.

I realise that `singularity --help` exists, but it's very common and useful to have a command reference online as well.

multimeric:

This PR adds some RST files and commands to the Makefile to automatically generate documentation for the Singularity CLI. This will give users a way to browse and link others to the command line documentation online. This is implemented through adding Singularity as a submodule.

- Fixes #73

## Example Screenshot
image 1


GodloveD:
Heya @TMiguelT !  Thanks very much for the contribution.  At first glance it looks like a great addition.  Someone will have a closer look at it asap.

devprincess:
Looks good to me, thank you for the contribution @TMiguelT :+1:

GodloveD:
Heya @TMiguelT, I had a little time to take a closer look at this.  Thanks again for this valuable contribution!  I'm willing to approve as is, but I'm wondering about something.  I found that it was a little cumbersome to build the docs after this change (since it also implies building Singularity itself).  I wonder if you know can think of a method for disabling this feature when one is simply interested in doing a quick build and checking the results.  In fact, it would be great if that were the default, and the user only got a *full* build if they explicitly asked for it.  What do you think?

multimeric:
Yes so I could tweak the makefile to have different targets, some that require the CLI docs and some that don't. Or perhaps add a parameter to the makefile that turns on or off CLI doc building. But the thing is, I kind of want the default to involve building them, since they need to now be a core part of the documentation. I don't want people accidentally forgetting them and excluding them from a documentation release.

GodloveD:
OK.  I think I can live with that as a sensible default.  Could you also add a line or two to the README.md so that contributors can see how to build easily?  My primary concern is making sure that the docs are easy to build so it is not an unnecessary obstacle for potential contributors.

multimeric:
Yep okay, so if I added a parameter such that you could `make html NO_CLI=1` or something, do you think that would be satisfactory? And then update the README accordingly

multimeric:
Okay @GodloveD, I've added a Makefile flag, so you can `make html SKIPCLI=1` and it will ignore everything Singularity related. I've also updated the README to explain this whole PR.

GodloveD:
Once again, I'm sorry that it took me a while to get to this @TMiguelT.  I've been super busy.  I've tested this and it looks like it works as advertised!

GodloveD (approved):

multimeric:
Great, thank you!

PR #98, opened 2019-02-18
Updated default bind points

### LINKED ISSUE #88: Document automatic binds
There should be documentation somewhere (if there is not already) specifying what directories and files are automatically bind mounted by default with the default `singularity.conf` file.

WestleyK:

This PR adds all the default bind points.

- Fixes #88

### DIFF
--- a/bind_paths_and_mounts.rst
+++ b/bind_paths_and_mounts.rst
@@ -32,14 +32,14 @@ The system administrator has the ability to define what bind paths will be
 included automatically inside each container. Some bind paths are automatically
 derived (e.g. a user’s home directory) and some are statically defined (e.g.
 bind paths in the Singularity configuration file). In the default
-configuration, the directories ``$HOME`` , ``/tmp:/tmp`` , ``/proc:/proc`` , ``/sys:/sys`` ,
+configuration, the system default bind points are ``$HOME`` , ``/tmp:/tmp`` , ``/proc:/proc`` , ``/sys:/sys`` ,
 ``/dev:/dev``, ``/proc:/proc``, ``/sys:/sys``, ``/dev:/dev``, ``/tmp:/tmp``, ``/var/tmp:/var/tmp``, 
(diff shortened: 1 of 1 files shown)


bind_paths_and_mounts.rst
```diff
@@ -32,8 +32,10 @@ The system administrator has the ability to define what bind paths will be
 included automatically inside each container. Some bind paths are automatically
 derived (e.g. a user’s home directory) and some are statically defined (e.g.
 bind paths in the Singularity configuration file). In the default
-configuration, the directories ``$HOME`` , ``/tmp`` , ``/proc`` , ``/sys`` ,
-``/dev``, and ``$PWD`` are among the system-defined bind paths.
+configuration, the directories ``$HOME`` , ``/tmp:/tmp`` , ``/proc:/proc`` , ``/sys:/sys`` ,
+``/dev:/dev``, ``/proc:/proc``, ``/sys:/sys``, ``/dev:/dev``, ``/tmp:/tmp``, ``/var/tmp:/var/tmp``, 
+``/etc/resolv.conf:/etc/resolv.conf``, ``/etc/passwd:/etc/passwd``, and ``$PWD``
+are the system-defined bind paths.
```
jchavezlavalle: maybe at this point the `<path from host>:<path from container>` notation isn't clear to the user since the `-b` flag is explained below.. could be good idea to add something to this paragraph like.. 

`Where the first part of the string before ":" is the path from the host and the second part is the path from the container`

jchavezlavalle (changes requested):

jchavezlavalle (approved):
lgtm! @WestleyK please make sure to rebase to master :+1:

GodloveD:
There seems to be some instances in which `/etc/localtime` is a default bind and some in which it is not.  Are other files also like this?  @WestleyK can you investigate before we merge?

GodloveD (approved):
I think this is fine to merge. I think it needs to be refined a bit more (maybe after merging) though.  For instance, `/etc/localtime` will also be automagically bind mounted into the container, but only if it exists in the container as a place to go right?

PR #104, opened 2019-02-19
Add warning on compressor issue on running an image on a different host

### LINKED ISSUE #95: mksquashfs compressor issues
It should be noted in documentation about the possibility of building an image on one host, but it *not* running on another host, because the squashfs tools does not support the compression format used when making the squashfs image in the SIF file.

For example, building an image on a host where the default compressor is `xz`, and then trying to run that image on a CentOS 6 node, where the only compressor available is `gzip`.

jchavezlavalle:

If the default compressor on a host is not the same on another host, the run command could fail because this compressor is not supported. This is explained now on the build a container section for 3.0. 

- Fixes #95

### DIFF
--- a/build_a_container.rst
+++ b/build_a_container.rst
@@ -159,7 +159,7 @@ requires root privileges.
 
 
 .. note::
-    Beware that it is possible to build an image on a host and then not running on a different host. This depends on
+    Beware that it is possible to build an image on a host and have the image not work on a different host. This could be because of
     the default compressor supported by the host. For example, when building an image on a host in which the default compressor
     is ``xz`` and then trying to run that image on a CentOS 6 node, where the only compressor available is ``gzip``.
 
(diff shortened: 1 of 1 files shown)


jchavezlavalle:
hey @jmstover could you review if this fixes #95 ? thanks !

build_a_container.rst
```diff
@@ -157,6 +157,12 @@ You can do so with the following command.
 The command requires ``sudo`` just as installing software on your local machine
 requires root privileges.
 
+
+.. note::
+    Beware that it is possible to build an image on a host and then not running on a different host. This depends on
```
jmstover: I think the wording here should be:

```
[...] on a host and have the image not work on a different host. This could be because of the default compressor [...]
```

jmstover (changes requested):

build_a_container.rst
jmstover: Other than the wording there, this looks good to me.

jmstover (approved):

PR #114, opened 2019-03-19
File for documenting the OCI runtime (WiP)

### LINKED ISSUE #87: Draft content on OCI runtime support
## Which Document page:

New section or add to the Docker / OCI interoperability section? 

### Expected results:

Documentation to support `singularity oci *` command - new to Singularity 3.1 

### Actual results:

What is the problem, what's misspelled, or what's missing?

syl-ilumb:

Introduce user-facing documentation for OCI runtime support introduced in version 3.1 of Singularity. 

- Fixes #87


oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+	2. The container's root filesystem - a referenced directory
+
+.. note::
+
+	Because the directory itself, i.e., ``/var/tmp/busybox`` is *not* part of the bundle, the mount point can be chosen arbitrarily. 
+
+The `filtered <https://github.com/stedolan/jq>`_ ``config.json`` file corresponding to the OCI mounted ``busybox_latest.sif`` container can be detailed as follows via ``$ sudo cat /var/tmp/busybox/config.json | jq``: 
+
+.. code-block:: json
```
sashayakovtseva, line 125: Maybe, shorten it somehow? or at least use compact formatting? this json is tooo long :)

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+OCI Runtime Support 
+===================
+
+.. _sec:oci_runtime_overview:
+
+--------
+Overview
+--------
+
+OCI is an acronym for the `Open Containers Initiative <https://www.opencontainers.org/>`_ - an independent organization whose mandate is to develop open standards relating to containerization. To date, standardization efforts have focused on container formats and runtimes. Singularity's compliance with respect to the OCI Image Specification is considered in detail :ref:`elsewhere <sec:oci_overview>`. It is compliance in the sense of the OCI Runtime Specification that is of concern here.
```
rherban: > It is compliance

either it's compliant or in compliance

rherban: I see what you're getting at, but i'm not thrilled about it.

oci_runtime.rst
rherban: Maybe just link to an example file in github? Does it help to have the full contents listed here?

syl-ilumb: Agreed! https://github.com/sylabs/singularity-userdocs/issues/121.

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+.. Check again after I create a bundle and container ... 
+
+.. sandbox???
+
+
+---------------------------------
+Unmounting OCI Filesystem Bundles
+---------------------------------
+
+Currently mounted OCI filesystem bundles can be easily verified:
```
rherban: Maybe we can skip this for the unmount section, since it's detailed above in mount

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+.. code-block:: none
+
+	$ sudo singularity oci umount /var/tmp/busybox
+
+To verify that the bundle has been unmounted, the ``df`` command can be issued again.  
+
+.. note:: 
+
+	The argument provided to ``oci umount`` above is the name of the bundle path, ``/var/tmp/busybox/rootfs``, as opposed to the mount point for the overlay filesystem, ``/var/tmp/busybox/rootfs``.
```
rherban: bundle path is /var/tmp/busybox

oci_runtime.rst
syl-ilumb: @rherban : Excellent suggestion! Will do.

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+--------
+Overview
+--------
+
+OCI is an acronym for the `Open Containers Initiative <https://www.opencontainers.org/>`_ - an independent organization whose mandate is to develop open standards relating to containerization. To date, standardization efforts have focused on container formats and runtimes. Singularity's compliance with respect to the OCI Image Specification is considered in detail :ref:`elsewhere <sec:oci_overview>`. It is compliance in the sense of the OCI Runtime Specification that is of concern here. 
+
+Briefly, compliance with respect to the OCI Runtime Specification is addressed in Singularity through the introduction of the ``oci`` command group. Although this command group can in principle be used to provide a runtime that supports end users, in this initial documentation effort, emphasis is placed upon interoperability with Kubernetes; more specifically, interoperability with Kubernetes via the `Singularity Container Runtime Interface <https://www.sylabs.io/guides/cri/1.0/user-guide/index.html>`_. 
+
+Owing to this restricted focus, a subset of the Singularity commands in this ``oci`` command group receive attention here; specifically:
```
sashayakovtseva: command in command group - repetition. is that ok?

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+	- Mounting and unmounting OCI filesystem bundles
+	- Creating OCI compliant container instances 
+	- Context for integration with Kubernetes via the Singularity CRI
```
sashayakovtseva: context you mean is not a subset of commands, so I'd remove it from list here, or rephrase

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+	Copying blob sha256:fc1a6b909f82ce4b72204198d49de3aaf757b3ab2bb823cb6e47c416b97c5985
+	 738.13 KiB / 738.13 KiB [==================================================] 0s
+	Copying config sha256:5fffaf1f2c1830a6a8cf90eb27c7a1a8476b8c49b4b6261a20d6257d031ce4f3
+	 575 B / 575 B [============================================================] 0s
+	Writing manifest to image destination
+	Storing signatures
+	INFO:    Creating SIF file...
+	INFO:    Build complete: busybox_latest.sif
+
+is one way to bootstrap creation of this image in SIF that *retains* a local copy. Additional approaches and details can be found in the section :ref:`Support for Docker and OCI <singularity-and-docker>`).
```
sashayakovtseva: hm, technically image is not created here but downloaded from cloud to local storage

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+	Copying config sha256:5fffaf1f2c1830a6a8cf90eb27c7a1a8476b8c49b4b6261a20d6257d031ce4f3
+	 575 B / 575 B [============================================================] 0s
+	Writing manifest to image destination
+	Storing signatures
+	INFO:    Creating SIF file...
+	INFO:    Build complete: busybox_latest.sif
+
+is one way to bootstrap creation of this image in SIF that *retains* a local copy. Additional approaches and details can be found in the section :ref:`Support for Docker and OCI <singularity-and-docker>`). 
+
+For the purpose of boostrapping the creation of an OCI compliant runtime, this SIF file can be mounted as follows:
```
sashayakovtseva: OCI runtime compliant container maybe? we do not create runtime here, only a bundle to start a container later

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+is one way to bootstrap creation of this image in SIF that *retains* a local copy. Additional approaches and details can be found in the section :ref:`Support for Docker and OCI <singularity-and-docker>`). 
+
+For the purpose of boostrapping the creation of an OCI compliant runtime, this SIF file can be mounted as follows: 
+
+.. code-block:: none 
+
+	$ sudo singularity oci mount ./busybox_latest.sif /var/tmp/busybox
+
+By issuing the ``mount`` command, the Singularity container runtime encapsulated in the SIF file ``busybox_latest.sif`` is mounted on the mount point ``/var/tmp/busybox`` as an ``overlay`` file system,
```
sashayakovtseva: Filesystem is encapsulated into a SIF, not runtime

sashayakovtseva: mounted on the mount point - unnecessary repetition imo

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+	$ sudo singularity oci mount ./busybox_latest.sif /var/tmp/busybox
+
+By issuing the ``mount`` command, the Singularity container runtime encapsulated in the SIF file ``busybox_latest.sif`` is mounted on the mount point ``/var/tmp/busybox`` as an ``overlay`` file system, 
+
+.. code-block:: none
+
+	$ sudo df -k
+	Filesystem                   1K-blocks    Used Available Use% Mounted on
+	udev                            475192       0    475192   0% /dev
```
sashayakovtseva, line 77: maybe remove all output except for target overlay? not to clutter things with your system's state

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+.. code-block:: none
+
+	$ sudo ls -ld /var/tmp/busybox
+	drwx------ 4 root root 4096 Apr  4 14:30 /var/tmp/busybox
+
+
+Content of an OCI Compliant Filesystem Bundle
+=============================================
+
+The *expected* contents of the mounted filesystem are as follows:
```
sashayakovtseva, line 97: why expected? there are no way contents changes unless  oci engine changes

sashayakovtseva: I'd remove expected here

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+	$ sudo ls -la /var/tmp/busybox
+	total 28
+	drwx------ 4 root root 4096 Apr  4 14:30 .
+	drwxrwxrwt 4 root root 4096 Apr  4 14:30 ..
+	-rw-rw-rw- 1 root root 9879 Apr  4 14:30 config.json
+	drwx------ 4 root root 4096 Apr  4 14:30 overlay
+	drwx------ 1 root root 4096 Apr  4 14:30 rootfs
+
+From the perspective of the `OCI runtime specification <https://github.com/opencontainers/runtime-spec/blob/master/bundle.md>`_, this content is expected because it prescribes a
```
sashayakovtseva, line 109: prescribes a "... a format" - double article

sashayakovtseva: is not expected, but required rather

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+	drwxrwxrwt 4 root root 4096 Apr  4 14:30 ..
+	-rw-rw-rw- 1 root root 9879 Apr  4 14:30 config.json
+	drwx------ 4 root root 4096 Apr  4 14:30 overlay
+	drwx------ 1 root root 4096 Apr  4 14:30 rootfs
+
+From the perspective of the `OCI runtime specification <https://github.com/opencontainers/runtime-spec/blob/master/bundle.md>`_, this content is expected because it prescribes a 
+
+	"... a format for encoding a container as a **filesystem bundle** - a set of files organized in a certain way, and containing all the necessary data and metadata for any compliant runtime to perform all standard operations against it." 
+
+Critical to compliance with the specification is the presence of the following *mandatory* artifacts residing locally in a single directory:
```
sashayakovtseva, line 113: So here you are describing OCI bundle spec, do we really want to do that? I mean, it is already described in OCI spec, we can put a link here...

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+	1. The ``config.json`` file - a file of configuration data that must reside in the root of the bundle directory under this name 
+
+	2. The container's root filesystem - a referenced directory
+
+.. note::
+
+	Because the directory itself, i.e., ``/var/tmp/busybox`` is *not* part of the bundle, the mount point can be chosen arbitrarily.
```
sashayakovtseva, line 121: IMO that is obvious and can be removed

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+	1. The ``config.json`` file - a file of configuration data that must reside in the root of the bundle directory under this name 
+
+	2. The container's root filesystem - a referenced directory
+
+.. note::
+
+	Because the directory itself, i.e., ``/var/tmp/busybox`` is *not* part of the bundle, the mount point can be chosen arbitrarily. 
+
+The `filtered <https://github.com/stedolan/jq>`_ ``config.json`` file corresponding to the OCI mounted ``busybox_latest.sif`` container can be detailed as follows via ``$ sudo cat /var/tmp/busybox/config.json | jq``:
```
sashayakovtseva, line 123: `jq` may not be installed on everyone's system

sashayakovtseva: what does filtered config.json means?

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+	            "modify_ldt"
+	          ],
+	          "action": "SCMP_ACT_ALLOW"
+	        }
+	      ]
+	    }
+	  }
+	}
+
+Furthermore, and through use of ``$ sudo cat /var/tmp/busybox/config.json | jq [.root.path]``, the property
```
sashayakovtseva, line 702: same as above, `jq` may not be installed, you should use standard commands here

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+.. code-block:: none
+
+	$ sudo ls /var/tmp/busybox/rootfs
+	bin  dev  environment  etc  home  proc	root  singularity  sys	tmp  usr  var
+
+.. note::
+
+	``environment`` and ``singularity`` above are symbolic links to the ``.singularity.d`` directory. 
+
+.. TODO Is the ``.singularity.d`` ignored in this case? Relates to the other quote I lifted ...
```
sashayakovtseva, line 721: it is not ignored, depends on how you will start the container

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+	``environment`` and ``singularity`` above are symbolic links to the ``.singularity.d`` directory. 
+
+.. TODO Is the ``.singularity.d`` ignored in this case? Relates to the other quote I lifted ... 
+
+	"The definition of a bundle is only concerned with how a container, and its configuration data, are stored on a local filesystem so that it can be consumed by a compliant runtime."
+
+Beyond ``root.path``, the ``config.json`` file includes a multitude of additional properties - for example:
+
+	- ``ociVersion`` - a mandatory property that identifies the version of the OCI runtime specification that the bundle is compliant with
```
sashayakovtseva, line 727: same as above, I think we don't need to describe OCI standard once and this can be omitted

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+.. TODO Is the ``.singularity.d`` ignored in this case? Relates to the other quote I lifted ... 
+
+	"The definition of a bundle is only concerned with how a container, and its configuration data, are stored on a local filesystem so that it can be consumed by a compliant runtime."
+
+Beyond ``root.path``, the ``config.json`` file includes a multitude of additional properties - for example:
+
+	- ``ociVersion`` - a mandatory property that identifies the version of the OCI runtime specification that the bundle is compliant with 
+
+	- ``process`` - an optional property that specifies the container process. When invoked via Singularity, subproperties such as ``args`` are populated by making use of the contents of the ``.singularity.d`` directory, e.g. via ``$ sudo cat /var/tmp/busybox/config.json | jq [.process.args]``:
```
sashayakovtseva, line 729: It is required when it comes to running containers... See https://github.com/opencontainers/runtime-spec/blob/master/config.md#process

sashayakovtseva: "When invoked via Singularity, subproperties such as ``args`` are populated by making use of the contents of the ``.singularity.d`` directory" - as far as I know that is not true, @cclerget may tell for sure

sashayakovtseva: and same for `jq`

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+		  [
+		    "/.singularity.d/actions/run"
+		  ]
+		]
+
+	where ``run`` equates to the :ref:`familiar runscript <sec:inspect_container_metadata>` for this container. 
+
+For a comprehensive discussion of all the ``config.json`` file properties, refer to the `implementation guide <https://github.com/opencontainers/runtime-spec/blob/master/config.md>`_. 
+
+Technically, the ``overlay`` directory was *not* content expected of an OCI compliant filesystem bundle. As detailed in the section dedicated to `Persistent Overlays <https://www.sylabs.io/guides/3.0/user-guide/persistent_overlays.html>`_, these directories allow for the introduction of
```
sashayakovtseva, line 743: "directory was *not* content expected" - I don't understand what are you trying to say here

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+For a comprehensive discussion of all the ``config.json`` file properties, refer to the `implementation guide <https://github.com/opencontainers/runtime-spec/blob/master/config.md>`_. 
+
+Technically, the ``overlay`` directory was *not* content expected of an OCI compliant filesystem bundle. As detailed in the section dedicated to `Persistent Overlays <https://www.sylabs.io/guides/3.0/user-guide/persistent_overlays.html>`_, these directories allow for the introduction of 
+a writable file system on an otherwise immutable read-only container; thus they permit the illusion of read-write access.
+
+.. TODO Need to ensure that what's written above is correct 
+
+.. note::
+
+	SIF is stated to be an extensible format capable of encasulating the entire container runtime in a single file. By encapsulating a filesystem bundle that conforms with the OCI runtime specification, the extensibility of SIF is demonstrably evident.
```
sashayakovtseva: container filesystem, not runtime. Unless I am missing something...

sashayakovtseva: demonstrably evident - tautology, isn't is?

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+
+------------------------------------------
+Creating OCI Compliant Container Instances 
+------------------------------------------
+
+SIF files encapsulate filesystem bundles that conform with the OCI runtime specification. By 'OCI mounting' a SIF file (see above), this encapsulated filesystem bundle is exposed. Once exposed, the filesystem bundle can be used to bootstrap the creation of an OCI compliant container instance as follows:
```
sashayakovtseva: Well, SIF doesn't encapsulate OCI bundles, but can be extracted to it rather. Or bundles can be created based on it, whatever

oci_runtime.rst
sashayakovtseva: +1

oci_runtime.rst
sashayakovtseva: +1

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+	"id": "busybox1",
+	"status": "created",
+	"pid": 6578,
+	"bundle": "/var/tmp/busybox",
+	"createdAt": 1554389921452964253,
+	"attachSocket": "/var/run/singularity/instances/root/busybox1/attach.sock",
+	"controlSocket": "/var/run/singularity/instances/root/busybox1/control.sock"
+	}
+
+.. TODO Confirmm the above is OCI standards compliant ^^^
```
sashayakovtseva: https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#state

sashayakovtseva (changes requested):

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+.. _sec:oci_runtime_overview:
+
+--------
+Overview
+--------
+
+OCI is an acronym for the `Open Containers Initiative <https://www.opencontainers.org/>`_ - an independent organization whose mandate is to develop open standards relating to containerization. To date, standardization efforts have focused on container formats and runtimes. Singularity's compliance with respect to the OCI Image Specification is considered in detail :ref:`elsewhere <sec:oci_overview>`. It is compliance in the sense of the OCI Runtime Specification that is of concern here. 
+
+Briefly, compliance with respect to the OCI Runtime Specification is addressed in Singularity through the introduction of the ``oci`` command group. Although this command group can in principle be used to provide a runtime that supports end users, in this initial documentation effort, emphasis is placed upon interoperability with Kubernetes; more specifically, interoperability with Kubernetes via the `Singularity Container Runtime Interface <https://www.sylabs.io/guides/cri/1.0/user-guide/index.html>`_.
```
GodloveD: `Although this command group can, in principle, be` <-- add commas like so.

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+Briefly, compliance with respect to the OCI Runtime Specification is addressed in Singularity through the introduction of the ``oci`` command group. Although this command group can in principle be used to provide a runtime that supports end users, in this initial documentation effort, emphasis is placed upon interoperability with Kubernetes; more specifically, interoperability with Kubernetes via the `Singularity Container Runtime Interface <https://www.sylabs.io/guides/cri/1.0/user-guide/index.html>`_. 
+
+Owing to this restricted focus, a subset of the Singularity commands in this ``oci`` command group receive attention here; specifically:
+
+	- Mounting and unmounting OCI filesystem bundles
+	- Creating OCI compliant container instances
```
GodloveD, line 24: - ~Creating~ Exposing SIF files to the host file system as OCI compliant container instances 
^ Is this what you mean?

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+Briefly, compliance with respect to the OCI Runtime Specification is addressed in Singularity through the introduction of the ``oci`` command group. Although this command group can in principle be used to provide a runtime that supports end users, in this initial documentation effort, emphasis is placed upon interoperability with Kubernetes; more specifically, interoperability with Kubernetes via the `Singularity Container Runtime Interface <https://www.sylabs.io/guides/cri/1.0/user-guide/index.html>`_. 
+
+Owing to this restricted focus, a subset of the Singularity commands in this ``oci`` command group receive attention here; specifically:
+
+	- Mounting and unmounting OCI filesystem bundles
+	- Creating OCI compliant container instances 
+	- Context for integration with Kubernetes via the Singularity CRI 
+
+Subsequent revisions of this documentation are anticipated to provide more complete coverage of this command group.
```
GodloveD: I'd just delete this.  Sounds a bit apologetic and there is no need top apologize for this well-written doc imho.

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+------------------------------
+
+Mounting an OCI Filesystem Bundle
+=================================
+
+For the purpose of illustration, use of `BusyBox <https://busybox.net/about.html>`_ will be made here. 
+
+Suppose the Singularity Image Format (SIF) file ``busybox_latest.sif`` exists locally. (Recall that
+
+.. code-block:: none
```
GodloveD, line 49: imho it is bad form to interrupt a sentence with a code block like this.  I think it should be rewritten so that it does not interrupt the flow like this.

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+.. TODO Compliance testing/validation  - need to document ... https://github.com/opencontainers/runtime-tools
+
+------------------------------
+Mounted OCI Filesystem Bundles
+------------------------------
+
+Mounting an OCI Filesystem Bundle
+=================================
+
+For the purpose of illustration, use of `BusyBox <https://busybox.net/about.html>`_ will be made here.
```
GodloveD: How about
`We'll uses busybox here for purposes of illustration.`
So much easier to parse imho.

oci_runtime.rst
GodloveD: Well, technically, a SIF is _created_ from the layers that are downloaded.  But the local copy that is retained is not a SIF.  It's the layers.  So, there is some nuance here that is not quite captured by the text I think.

oci_runtime.rst
GodloveD: The use of the term `bootstrapping` here is a bit confusing.  It's just repackaging a container from one format into another (and losing some info in the process).  It will also be doubly confusing for folks who have been using Singularity for a while because `bootstrap` used to be a command.  And it used to do something totally different.

oci_runtime.rst
GodloveD: Echoing @sashayakovtseva here.  It's a root filesystem.  Not a runtime.

oci_runtime.rst
GodloveD: Just put the `$ sudo cat /var/tmp/busybox/config.json | jq` command into the codeblock as well.  Don't separate it from the output I think.

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+	    }
+	  }
+	}
+
+Furthermore, and through use of ``$ sudo cat /var/tmp/busybox/config.json | jq [.root.path]``, the property
+
+.. code-block:: json
+
+	[
+  	"/var/tmp/busybox/rootfs"
```
GodloveD: Indentation needed on this line.

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+.. code-block:: none
+
+	$ sudo singularity oci create -b /var/tmp/busybox busybox1
+
+In this example, the filesystem bundle is located in the directory ``/var/tmp/busybox`` - i.e., the mount point identified above with respect to 'OCI mounting'. The ``config.json`` file, along with the ``rootfs`` and ``overlay`` filesystems, are all employed in the bootstrap process. The instance is named ``busybox1`` in this example. 
+
+.. note::
+
+	The outcome of this creation request is truly a container **instance**. Multiple instances of the same container can easily be created by simply changing the name of the instance upon subsequent invocation requests.
```
GodloveD, line 773: Is it really an instance though?  I thought we determined it was not. As in, you cannot join this instance with the shell command.

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+	$ sudo singularity oci create -b /var/tmp/busybox busybox1
+
+In this example, the filesystem bundle is located in the directory ``/var/tmp/busybox`` - i.e., the mount point identified above with respect to 'OCI mounting'. The ``config.json`` file, along with the ``rootfs`` and ``overlay`` filesystems, are all employed in the bootstrap process. The instance is named ``busybox1`` in this example. 
+
+.. note::
+
+	The outcome of this creation request is truly a container **instance**. Multiple instances of the same container can easily be created by simply changing the name of the instance upon subsequent invocation requests. 
+
+The ``state`` of the container instance can be determined via ``$ sudo singularity oci state busybox``:
```
GodloveD: ~busybox~ busybox1

oci_runtime.rst
GodloveD: Also the phrasing is confusing.  It kindof sounds like this can be verified currently, but we will remove this command in the future or something.  I agree to just remove it.

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+	The argument provided to ``oci umount`` above is the name of the bundle path, ``/var/tmp/busybox/rootfs``, as opposed to the mount point for the overlay filesystem, ``/var/tmp/busybox/rootfs``. 
+
+.. TODO Affect on running instances ... 
+
+
+
+.. https://www.sylabs.io/guides/cri/1.0/user-guide/installation.html?highlight=oci#install-dependencies 
+
+----------------------
+Kubernetes Integration
```
GodloveD, line 839: It seems like you are making a case for using K8s with Singularity.  That's good, but not sure it belongs in a technical manual.  I would suggest leaving it here for now, writing a blogpost (if one doesn't already exist) and then link it here and remove the duplicate content.

GodloveD (changes requested):
Heya @syl-ilumb thank you for this excellent page!  I think it need some tweaks, but then it should be good to go.  😸

oci_runtime.rst
cclerget: @GodloveD @syl-ilumb A SIF image is created from OCI image layers and both OCI layers and SIF image are cached locally

oci_runtime.rst
cclerget: `OCI image runtime bundle` ?

oci_runtime.rst
cclerget: In `singularity mount` context, there is two possibilities actually:

1. SIF image contains an OCI image configuration (eg: bootstrapped from docker or OCI registry) so singularity populates `args` with entrypoint/cmd defines in OCI image configuration
2. SIF image doesn't contain OCI image configuration (eg: bootstrapped from scratch or library) so singularity populates `args` with `/.singularity.d/actions/run` as default entrypoint

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+	"pid": 6578,
+	"bundle": "/var/tmp/busybox",
+	"createdAt": 1554389921452964253,
+	"attachSocket": "/var/run/singularity/instances/root/busybox1/attach.sock",
+	"controlSocket": "/var/run/singularity/instances/root/busybox1/control.sock"
+	}
+
+.. TODO Confirmm the above is OCI standards compliant ^^^ 
+
+Whereas the above is provided via the OCI command group, container instances created in this fashion are still 'known' to Singularity - for example:
```
cclerget: This is not true anymore and has been fixed https://github.com/sylabs/singularity/issues/3103. To avoid confusion because `singularity instance` can't interact with OCI containers and vice versa. This block can be removed

oci_runtime.rst
```diff
@@ -0,0 +1,899 @@
...
+
+	$ sudo singularity instance list 
+	INSTANCE NAME    PID      IMAGE
+	busybox1         6578     /var/tmp/busybox/var/tmp/busybox/rootfs
+
+Because these instances are owned by ``root``, use of ``sudo`` is *required* here. 
+
+.. note::
+
+	"A container instance is a persistent and isolated version of the container image that runs in the background." The section on :ref:`Running Services <running_services>` details Singularity's ``instance`` command, and provides various examples.
```
cclerget: Regarding previous comment on `instance`, maybe rephrase this into something like, OCI containers have some similarities with Singularity instances. Or maybe remove it ?

cclerget (changes requested):

oci_runtime.rst
sashayakovtseva: maybe we should say that upon bundle creation `args` will be filled by singularity automatically? The confusing part for me was `When invoked via Singularity`.

oci_runtime.rst
sashayakovtseva: agree

oci_runtime.rst
sashayakovtseva: `and losing some info in the process` - what info do we lose? OCI `config.json` is now embedded into a SIF so we don't lose anything

oci_runtime.rst
sashayakovtseva: Not singularity instance, but a container object, aka container instance. And we can shell into that container

oci_runtime.rst
syl-ilumb: @GodloveD : This is a code block formatted via the `json` pygment; if I place `$ sudo cat /var/tmp/busybox/config.json | jq` inside it, it'll also get that `json` pygmentation ...

syl-ilumb: @rherban: Great suggestion! Can you please assist?

syl-ilumb: @GodloveD: You're looking a tad pale today ... perhaps you could use a little of that old `json` pygmentation ...

sashayakovtseva: you can use gist: https://gist.github.com/

oci_runtime.rst
syl-ilumb: @GwendolynK: Any thoughts on this?

oci_runtime.rst
syl-ilumb: @GodloveD: From my days as a sysadmin, sanity checks were standard operating procedure; hence, the 'verification'. I have, however, removed this content as requested.

oci_runtime.rst
syl-ilumb: Apparently, I don't read what I write. Thank you! Fixed.

oci_runtime.rst
syl-ilumb: Will leave open the comment below re: the length of the `config.json` snippet ...

oci_runtime.rst
syl-ilumb: Yes. No. I cleaned this up. Thx @sashayakovtseva.

oci_runtime.rst
syl-ilumb: Addressed - thx!

oci_runtime.rst
syl-ilumb: According to https://www.sylabs.io/2018/03/sif-containing-your-containers/:

> One of Singularity’s design concepts in architecting a container solution is to  store a container as a single run-time file. 

Given that I can execute a container directly, e.g., `./lolcow.sif`, I'm still thinking that:

> the Singularity container runtime encapsulated in the SIF file 

What am I missing?

syl-ilumb: @sashayakovtseva: Removed redundant use of "mount" and "mount point" ... Thx

oci_runtime.rst
syl-ilumb: As above ... Perhaps I'm missing something ... Addressed tautological statement ;-)

oci_runtime.rst
syl-ilumb: Conferred with @cclerget. Rewrote - be sure to see the note that has the details.

oci_runtime.rst
rherban: PR131 adds an example json right in the repo.

oci_runtime.rst
syl-ilumb: Added - thank you for that pointer @sashayakovtseva!

oci_runtime.rst
syl-ilumb: Commas added :-)

oci_runtime.rst
syl-ilumb: Is revealing OK, @GodloveD?

oci_runtime.rst
syl-ilumb: That's the Canuck in me - deleted! Thx!

oci_runtime.rst
syl-ilumb: Used to do this all the time with equations in my past life as a geophysicist ... Nonetheless, majority rules! No more interruption!

oci_runtime.rst
syl-ilumb: ... but I get paid by the word ... by the word ... by the word ... OK, done.

oci_runtime.rst
syl-ilumb: ... an OCI container instance ... in other words, an executing container created from the _same_ OCI fs bundle ... isn't that an instance?

oci_runtime.rst
syl-ilumb: Yes - nice catch - thx @GodloveD. Fixed.

oci_runtime.rst
syl-ilumb: Fair point. Will address as suggested.

oci_runtime.rst
syl-ilumb: Block removed

oci_runtime.rst
syl-ilumb: Removed

oci_runtime.rst
syl-ilumb: Agreed. Moved. Thanks @sashayakovtseva.

oci_runtime.rst
syl-ilumb: Changed to container - thx @sashayakovtseva

oci_runtime.rst
syl-ilumb: Agreed. Fixed.

syl-ilumb:
Addressed https://github.com/sylabs/singularity-userdocs/pull/114#discussion_r272915745 - thx @cclerget

oci_runtime.rst
GodloveD: In that context, run-time is being used as an adjective.  Like a file that can be used at runtime.  In this context you are using run-time as a noun.  That is not correct.  The singularity run-time is what gets invoked when you use the singularity action commands `run, exec, shell, oci start` etc.

It's also important to differentiate between the two, because OCI specifies a spec for the "runtime" and "image format".  And you are talking about a runtime being encapsulated within the SIF image format.  Which it's not.  So it's quite confusing.  

You should really just stick to calling it a "root filesystem" I think.

oci_runtime.rst
GwendolynK-zz: In this case, saying "compliant" (instead of "is compliance") would be the better choice. 
("It is compliant in the sense of the OCI Specification that is of concern here.")

Another optional edit to make it a little more direct:
"It is compliant with the OCI Runtime Specification that is of concern here."

oci_runtime.rst
syl-ilumb: I am now referring to this as a root fs. Upon re-reading what I wrote initially, I now appreciate the confusion caused ... Thx @GodloveD & @sashayakovtseva.

oci_runtime.rst
syl-ilumb: Clarified - thx

oci_runtime.rst
syl-ilumb: Improved wording

PR #130, opened 2019-04-05
Tidied up installation page

rherban:

Clarified commands for installation.
Re-ordered so installation from release is higher focus than from source.
Added wget dependency.

### DIFF
--- a/installation.rst
+++ b/installation.rst
@@ -56,6 +56,7 @@ similar with other package managers.
         libgpgme11-dev \
         squashfs-tools \
         libseccomp-dev \
+        wget \
         pkg-config
 
 ``yum``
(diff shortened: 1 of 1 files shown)


installation.rst
```diff
@@ -107,6 +108,22 @@ for dependency resolution.
...
+==================================
+Install Singularity from a release
+==================================
+
+You can install Singularity from one of our releases. To see a full list, visit <https://github.com/sylabs/singularity/releases>.
+After that you can just run the following commands to proceed with the installation.
+
+.. code-block:: none
+
+    $ export VERSION=3.0.3 && # adjust this as necessary \
```
GodloveD: When these changes are made to the 3.1 branch make sure to update all versions to 3.1.1.  It would be really nice if there were some way in Sphynx to set a variable so that all occurrences of something like a version number could be updated.  We had something like that with the old jekyl pages.

installation.rst
```diff
@@ -107,6 +108,22 @@ for dependency resolution.
 
     $ go get -u github.com/golang/dep/cmd/dep
 
+==================================
+Install Singularity from a release
```
GodloveD: This heading is not really accurate b/c we are not yet installing.  Maybe `Download Singularity from a release` instead?

installation.rst
```diff
@@ -107,6 +108,22 @@ for dependency resolution.
...
+
+    $ export VERSION=3.0.3 && # adjust this as necessary \
+        mkdir -p $GOPATH/src/github.com/sylabs && \
+        cd $GOPATH/src/github.com/sylabs && \
+        wget https://github.com/sylabs/singularity/releases/download/v${VERSION}/singularity-${VERSION}.tar.gz && \
+        tar -xzf singularity-${VERSION}.tar.gz && \
+        cd singularity 
+
 ===================
 Install from source
```
GodloveD: Same thing here (and elsewhere if applicable) maybe change Install --> Download

GodloveD (changes requested):
Suggesting a few tweaks.

installation.rst
rherban: versions updated to 3.1.1.

jchavezlavalle (approved):

GodloveD (approved):

PR #131, opened 2019-04-08
Example config.json

rherban:

Adding config.json example file to be linked from OCI doc page.


syl-ilumb (approved):
A welcome extraction - thank you! Can't wait for the URL.

jchavezlavalle (approved):

GodloveD (changes requested):
I'm confused by this PR.  There are a bunch of changes on `installation.rst` that look like they are actually on another PR?  Is this a mistake?

Also, I think you should be able to stick a relative URI in right now for this page right?

GodloveD:
OK, now the PR is cleaner, but I'm still confused about the intention.  I _think_ that the intention is to remove the long cumbersome json from the OCI runtime page and make it a link instead.

But by itself this PR doesn't really do anything.  There is just this .json hanging out in the repo.  So, if I'm right and the intention is to remove the json from the OCI runtime page, here are the things that still should be done:

- change the json to an rst and wrap the actual text in a code block with formatting for json
- remove the json text from the OCI runtime page
- stick a link in there instead so that the user can click and be taken to the json page

GodloveD:
sounds like this is already fixed by having a vertical scroll bar.

PR #145, opened 2019-04-12
Update `singularity keys` to `singularity key`

### LINKED ISSUE #144: 3.1 docs point to deprecated full-form "keys" command
## Which Document page:

- https://www.sylabs.io/guides/3.1/user-guide/quick_start.html#overview-of-the-singularity-interface
- https://www.sylabs.io/guides/3.1/user-guide/signNverify.html

### Expected results:

The command `singularity key` used.

### Actual results:

The current docs still reference `singularity keys` which is the old, deprecated full-form version of the command.

phpHavok:

This is a minor PR which updates the `singularity keys` command to `singularity key` which is the new standardized form. The old command still works, but it's probably best from a QA standpoint not to advertise it anymore. The `singularity help` output is also updated, which includes the fix plus a few other commands that now exist in Singularity 3.1 but were missing from the docs.

- Fixes #144

### DIFF
--- a/quick_start.rst
+++ b/quick_start.rst
@@ -155,27 +155,30 @@ follows:
       -s, --silent             only print errors
       -t, --tokenfile string   path to the file holding your sylabs
                                authentication token (default
-                               "/home/david/.singularity/sylabs-token")
+                               "/home/jacob/.singularity/sylabs-token")
       -v, --verbose            print additional information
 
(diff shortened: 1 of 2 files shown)


jchavezlavalle:
looks good thanks @phpHavok :+1:

jchavezlavalle (approved):

PR #160, opened 2019-05-01
Add multi-stage build

### LINKED ISSUE #143: must document multi-stage builds
Added the `Stage: <name>` keyword to the definition file header and the from `<stage name>` option/argument pair to the %files section to support multistage builds.  Must be documented with examples.

rherban:

Adding documentation for multi-stage builds.  

- Fixes #143

### DIFF
--- a/definition_files.rst
+++ b/definition_files.rst
@@ -508,7 +508,7 @@ Singularity 3.2 introduces multi-stage builds where one environment can be used 
       func main() {
         fmt.Printf("Hello World!\n")
       }
-      EOF
+    EOF
 
       go build -o hello hello.go
 
--- a/definition_files.rst
+++ b/definition_files.rst
@@ -54,6 +54,10 @@ will pull a container from the `Container Library
 bootstrap agent will pull docker layers from `Docker Hub
 <https://hub.docker.com/>`_ as a base OS to start your image.
 
+Starting with Singularity 3.2, the ``Bootstrap`` keyword needs to be the first 
+entry in the header section.  This breaks compatibility with older versions 
+that allow the parameters of the header to appear in any order.
+
 Depending on the value assigned to ``Bootstrap``, other keywords may also be
 valid in the header. For example, when using the ``library`` bootstrap agent,
 the ``From`` keyword becomes valid. Observe the following example for building a
@@ -523,7 +527,7 @@ Singularity 3.2 introduces multi-stage builds where one environment can be used 
     %files from build
       /root/hello /bin/hello
 
-The stage names are arbitrary and files can be copied from any stage to another.
+The names of stages are arbitrary. Files can only be copied from stages declared before the current stage in the definition. E.g., the "devel" stage in the above definition cannot copy files from the "final" stage, but the "final" stage can copy files from the "devel" stage.
 
 ----
 Apps
--- a/definition_files.rst
+++ b/definition_files.rst
@@ -472,6 +472,52 @@ After building the help can be displayed like so:
         This is a demo container used to illustrate a def file that uses all
         supported sections.
 
+------------------
+Multi-Stage Builds
+------------------
+
+Singularity 3.2 introduces multi-stage builds where one environment can be used for compilation, then the resulting binary can be copied into a final environment.  This allows a slimmer final image that does not require the entire development stack.
+
+.. code-block:: singularity
+
+    Bootstrap: docker
+    From: golang:1.12.3-alpine3.9
+    Stage: build
+
+    %post
+            # prep environment
+        export PATH="/go/bin:/usr/local/go/bin:$PATH"
+        export HOME="/root"
+        cd /root
+
+        # insert source code, could also be copied from host with %files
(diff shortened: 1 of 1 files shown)


definition_files.rst
```diff
@@ -472,6 +472,52 @@ After building the help can be displayed like so:
         This is a demo container used to illustrate a def file that uses all
         supported sections.
 
+------------------
+Multi-Stage Builds
+------------------
+
+Singularity 3.2 introduces multi-stage builds where one environment can be used for compilation, then the resulting binary can be copied into a final environment.  This allows a slimmer final image that does not require the entire development stack.
+
+.. code-block:: singularity
```
jchavezlavalle, line 495: Maybe check the indentation around the def file? thanks @rherban

rherban:
Add comment to files to copy from another image.
add 'stage' to the header section

rherban:
@GodloveD @devprincess your comments have been addressed, please re-review when you can.

jchavezlavalle (approved):

jchavezlavalle:
lgtm, thanks :+1: @rherban

definition_files.rst
```diff
@@ -472,6 +482,49 @@ After building the help can be displayed like so:
...
+    # Install binary into final image
+    Bootstrap: library
+    From: alpine:3.9
+    Stage: final
+
+    # install binary from stage one
+    %files from build
+      /root/hello /bin/hello
+
+The stage names are arbitrary and files can be copied from any stage to another.
```
ikaneshiro: We should change this line to something like:
`The names of stages are arbitrary. Files can only be copied from stages declared before the current stage in the definition. E.g., the "devel" stage in the above definition cannot copy files from the "final" stage, but the "final" stage can copy files from the "devel" stage.`

ikaneshiro (changes requested):
We should add something easily seen at the very top of this page that says:
`In order to facilitate the new multi-stage build feature, the definition parser in 3.2+ now requires "Bootstrap" to be the first parameter of header. This breaks compatibility with previous definition parsers that allow the parameters of the header to appear in any order.`

rherban:
> We should add something easily seen at the very top of this page that says:
> `In order to facilitate the new multi-stage build feature, the definition parser in 3.2+ now requires "Bootstrap" to be the first parameter of header. This breaks compatibility with previous definition parsers that allow the parameters of the header to appear in any order.`

To clarify @ikaneshiro, Bootstrap has to be first regardless if you're using multi-stage?  Did that get called out as a breaking change?

ikaneshiro:
@rherban Yes for any build with a header, and this is a known breaking change and was encountered and discussed here https://github.com/sylabs/singularity/pull/3377

rherban:
I didn't see anything in the release notes on this and agree that something should have been said to users aren't left wondering why things changed after the upgrade.  I have updated based on your others comments @ikaneshiro .

ikaneshiro:
@rherban It was added to the changelog, but it looks like it wasn't mentioned in the release announcement which maybe it should have been. Understandable though since the security issue was so important it was probably the last thing on @GodloveD's mind

GodloveD:
I'm actually not seeing this in the `CHANGELOG.md` either 😞.  But yeah, we totally need to make users aware of it.

ikaneshiro:
@GodloveD https://github.com/sylabs/singularity/blob/master/CHANGELOG.md#new-features--functionalities Its there, sub bullet of the second bullet point

definition_files.rst
```diff
@@ -472,6 +486,49 @@ After building the help can be displayed like so:
...
+
+      # insert source code, could also be copied from host with %files
+      cat << EOF > hello.go
+      package main
+      import "fmt"
+
+      func main() {
+        fmt.Printf("Hello World!\n")
+      }
+    EOF
```
GodloveD, line 515: The whitespace in front of this EOF breaks the def file.  Can you remove?

GodloveD (changes requested):

GodloveD:
Oh hot damn @ikaneshiro !!  Great!  I missed that.

GodloveD (approved):
:shipit:

definition_files.rst
```diff
@@ -472,6 +486,49 @@ After building the help can be displayed like so:
...
+    # Install binary into final image
+    Bootstrap: library
+    From: alpine:3.9
+    Stage: final
+
+    # install binary from stage one
+    %files from build
+      /root/hello /bin/hello
+
+The names of stages are arbitrary. Files can only be copied from stages declared before the current stage in the definition. E.g., the "devel" stage in the above definition cannot copy files from the "final" stage, but the "final" stage can copy files from the "devel" stage.
```
ikaneshiro: Can you quote the quotted text using markdown syntax so it looks like `devel`, `final`, etc. I used double quotes in my last comment since github comments also interpret markdown and it would have gotten confused.

ikaneshiro (changes requested):
One nitpick, lgtm otherwise

ikaneshiro (approved):

PR #172, opened 2019-05-16
simplify installation docs a bit

GodloveD:
Plz make sure these procedures still work and that I didn't break nothings!

Screen Shot 2019-05-15 at 10 03 08 PM


jchavezlavalle (approved):
awesome we totally needed these, thanks @GodloveD

Issue #188, opened 2019-06-04
Need to update Quickstart guide for v3.2


sushma-4:
## Which Document page:

Quickstart

### Expected results:

Installation of Singularity does not require setting of Go paths and compiling Singularity in that directory.

PR #191, opened 2019-06-10
Updated the Quick-start guide

sushma-4:

Updated the quick guide so that the installation is described in the simplest way possible - compiling outside of GO etc. Also changed some command syntax supported by newer versions, updated the hyperlinks and interlinks to different sections of the docs.

- Fixes #
#188

### DIFF
--- a/installation.rst
+++ b/installation.rst
@@ -254,11 +254,9 @@ of the default ``/var``) you can do the following:
 Remove an old version
 =====================
 
-When you run ``sudo make install``, the command lists files as they are
-installed. They must all be removed in order to completely remove Singularity.
-
-For example, in a standard installation of Singularity 3.0.1 and beyond (when
-building from source) you must remove all of these files and directories to
+In a standard installation of Singularity 3.0.1 and beyond (when
+building from source), the command ``sudo make install`` lists all the files as
+they are installed. You must remove all of these files and directories to
 completely remove Singularity.
 
 Obviously, this list of files may differ depending on how you install
--- a/installation.rst
+++ b/installation.rst
@@ -6,7 +6,7 @@ Installation
 .. _sec:installation:
 
 This document will guide you through the process of installing Singularity >=
-{InstallationVersion} via several different methods. (For instructions on 
+{InstallationVersion} via several different methods. (For instructions on
 installing earlier versions of Singularity please see `earlier versions of the
 docs <https://www.sylabs.io/docs/>`_.)
 
@@ -16,15 +16,15 @@ Overview
 
 Singularity runs on Linux natively and can also be run on Windows and Mac
 through virtual machines (VMs). Here we cover several different methods of
-installing Singularity (>={InstallationVersion}) on Linux and also give methods 
-for downloading and running VMs with singularity pre-installed from `Vagrant 
+installing Singularity (>={InstallationVersion}) on Linux and also give methods
+for downloading and running VMs with singularity pre-installed from `Vagrant
 Cloud <https://app.vagrantup.com/sylabs>`_.
 
 Install on Linux
 ----------------
 
-Linux is the only operating system that can support containers because of kernel 
-features like namespaces.  You can use these methods to install Singularity on 
+Linux is the only operating system that can support containers because of kernel
+features like namespaces.  You can use these methods to install Singularity on
 bare metal Linux or a Linux VM.
 
 ================
@@ -79,7 +79,7 @@ similar with other package managers.
 Install Go
 ----------
 
-Singularity v3 is written primarily in Go, and you will need Go installed to 
+Singularity v3 is written primarily in Go, and you will need Go installed to
 compile it from source.
 
 This is one of several ways to `install and configure Go
@@ -109,9 +109,9 @@ Then, set up your environment for Go.
 Download Singularity from a release
 ===================================
 
-You can download Singularity from one of the releases. To see a full list, visit 
-`the GitHub release page <https://github.com/sylabs/singularity/releases>`_. 
-After deciding on a release to install, you can run the following commands to 
+You can download Singularity from one of the releases. To see a full list, visit
+`the GitHub release page <https://github.com/sylabs/singularity/releases>`_.
+After deciding on a release to install, you can run the following commands to
 proceed with the installation.
 
 .. code-block:: none
@@ -127,7 +127,7 @@ Download from source
 
 The following commands will install Singularity from the `GitHub repo
 <https://github.com/sylabs/singularity>`_ to ``/usr/local``. This method will
-work for >=v{InstallationVersion}. To install an older tagged release see `older 
+work for >=v{InstallationVersion}. To install an older tagged release see `older
 versions of the docs <https://www.sylabs.io/docs/>`_.
 
 When installing from source, you can decide to install from either a **tag**, a
@@ -161,8 +161,8 @@ to generate a ``Makefile`` and then ``make`` is used to compile and install.
         make -C ./builddir && \
         sudo make -C ./builddir install
 
-By default Singularity will be installed in the ``/usr/local`` directory 
-hierarchy. You can specify a custom directory with the ``--prefix`` option, to 
+By default Singularity will be installed in the ``/usr/local`` directory
+hierarchy. You can specify a custom directory with the ``--prefix`` option, to
 ``mconfig`` like so:
 
 .. code-block:: none
@@ -214,8 +214,8 @@ directly from the `release tarball
     should be named `singularity-<version>.tar.gz`.
 
 After installing the :ref:`dependencies <install-dependencies>` and installing
-:ref:`Go <install-go>` as detailed above, you are ready download the tarball and
-build and install the RPM.
+:ref:`Go <install-go>` as detailed above, you are ready to download the tarball
+and build and install the RPM.
 
 .. code-block:: none
 
@@ -225,12 +225,12 @@ build and install the RPM.
         sudo rpm -ivh ~/rpmbuild/RPMS/x86_64/singularity-$VERSION-1.el7.x86_64.rpm && \
         rm -rf ~/rpmbuild singularity-$VERSION*.tar.gz
 
-If you encounter a failed dependency error for golang but installed it from 
+If you encounter a failed dependency error for golang but installed it from
 source, build with this command:
 
 .. code-block:: none
 
-    rpmbuild -tb --nodeps singularity-${VERSION}.tar.gz 
+    rpmbuild -tb --nodeps singularity-${VERSION}.tar.gz
 
 
 Options to ``mconfig`` can be passed using the familiar syntax to ``rpmbuild``.
@@ -257,8 +257,8 @@ Remove an old version
 When you run ``sudo make install``, the command lists files as they are
 installed. They must all be removed in order to completely remove Singularity.
 
-For example, in a standard installation of Singularity 3.0.1 and beyond (when 
-building from source) you must remove all of these files and directories to 
+For example, in a standard installation of Singularity 3.0.1 and beyond (when
+building from source) you must remove all of these files and directories to
 completely remove Singularity.
 
 Obviously, this list of files may differ depending on how you install
@@ -440,8 +440,8 @@ With Singularity Desktop for macOS (Alpha Preview):
 The disk image file is available `here <http://repo.sylabs.io/desktop/singularity-desktop-0.0.1alpha.dmg>`_.
 More information can be found `here <https://www.sylabs.io/singularity-desktop-macos/>`_.
 
-Singularity is also available via Vagrant (installable with 
-`Homebrew <https://brew.sh>`_ or manually) or with the Singularity Desktop for 
+Singularity is also available via Vagrant (installable with
+`Homebrew <https://brew.sh>`_ or manually) or with the Singularity Desktop for
 macOS (Alpha Preview).
 
 With Homebrew:
@@ -494,7 +494,7 @@ You can check the installed version of Singularity with the following:
 Of course, you can also start with a plain OS Vagrant box as a base and then
 install Singularity using one of the above methods for Linux.
 
-.. _installation-request:
+.. _installationrequest:
 
 Singularity on a shared resource
 --------------------------------
--- a/installation.rst
+++ b/installation.rst
@@ -17,7 +17,7 @@ Overview
 Singularity runs on Linux natively and can also be run on Windows and Mac
 through virtual machines (VMs). Here we cover several different methods of
 installing Singularity (>={InstallationVersion}) on Linux and also give methods
-for downloading and running VMs with singularity pre-installed from `Vagrant
+for downloading and running VMs with Singularity pre-installed from `Vagrant
 Cloud <https://app.vagrantup.com/sylabs>`_.
 
 Install on Linux
--- a/quick_start.rst
+++ b/quick_start.rst
@@ -85,7 +85,7 @@ Set the Environment variable ``PATH`` to point to Go:
 
 .. code-block:: none
 
-    $  echo 'export PATH=/usr/local/go/bin' >> ~/.bashrc && \
+    $  echo 'export PATH=/usr/local/go/bin:$PATH' >> ~/.bashrc && \
     source ~/.bashrc
 
 .. _download:
--- a/quick_start.rst
+++ b/quick_start.rst
@@ -27,7 +27,7 @@ You will need a Linux system to run Singularity.
 
 .. note::
      This is only a short manual for Quick Installation. For details on
-     different methods of Installation, versions, and RPM etc, please refer to
+     different methods of installation, versions, and RPM etc, please refer to
      the :ref:`Installation page <installation>`.
 
 Install system dependencies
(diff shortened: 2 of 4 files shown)


quick_start.rst
```diff
@@ -25,8 +25,10 @@ Quick Installation Steps
 
 You will need a Linux system to run Singularity.
 
-See the :ref:`installation page <installation>` for information about installing
-Singularity.
+.. note::
+     This is only a short manual for Quick Installation. For details on
+     different methods of Installtion, versions, and RPM etc, please refer to
```
jchavezlavalle: typo on word Installation

quick_start.rst
```diff
@@ -48,46 +50,52 @@ You must first install development libraries to your host. Assuming Ubuntu
...
 .. code-block:: none
 
-    $ export VERSION=1.12 OS=linux ARCH=amd64 && \
-        wget https://dl.google.com/go/go$VERSION.$OS-$ARCH.tar.gz && \
-        sudo tar -C /usr/local -xzvf go$VERSION.$OS-$ARCH.tar.gz && \
-        rm go$VERSION.$OS-$ARCH.tar.gz
+    $  sudo tar -C /usr/local -xzvf go1.12.5.linux-amd64.tar.gz
 
-Then, set up your environment for Go.
+Set the Environment variable PATH to point to Go:
```
jchavezlavalle: better use ``PATH`` notation with the `` quotes

jchavezlavalle (changes requested):
left some comments around @sushma-4 thanks!

quick_start.rst
```diff
@@ -48,46 +50,52 @@ You must first install development libraries to your host. Assuming Ubuntu
...
+    Note that ``squashfs-tools`` is only a dependency for commands that builds
+    images. The ``build`` command obviously relies on ``squashfs-tools``, but
+    other commands may do so as well if they are run using container images
     from Docker Hub for instance.
 
+There are 3 broad steps to installing Singularity:
+
+1. :ref:`Installing Go <install>`
+2. :ref:`Downloading Singularity <download>`
+3. :ref:`Compiling Singularity Binaries <compile>`
```
WestleyK: Source code maybe?

```suggestion
3. :ref:`Compiling Singularity Source Code <compile>`
```

quick_start.rst
```diff
@@ -48,46 +50,52 @@ You must first install development libraries to your host. Assuming Ubuntu
...
-Then, set up your environment for Go.
+Set the Environment variable ``PATH`` to point to Go:
 
 .. code-block:: none
 
-    $ echo 'export GOPATH=${HOME}/go' >> ~/.bashrc && \
-        echo 'export PATH=/usr/local/go/bin:${PATH}:${GOPATH}/bin' >> ~/.bashrc && \
-        source ~/.bashrc
+    $  echo 'export PATH=/usr/local/go/bin' >> ~/.bashrc && \
+    source ~/.bashrc
```
WestleyK: Why did you change this command?

quick_start.rst
```diff
@@ -97,6 +105,8 @@ proceed with the installation.
         tar -xzf singularity-${VERSION}.tar.gz && \
         cd singularity
 
+.. _compile:
+
 Compile the Singularity binary
```
WestleyK: Source code maybe?

```suggestion
Compile the Singularity source code
```

WestleyK (commented):
Looks good, couple suggestions/questions...

quick_start.rst
sushma-4: There was no need to set GOPATH, just making the PATH = /usr/local/go/bin was enough.. idk did you have to set the GOPATH too for your installation?

WestleyK: First, you dont want you PATH to only be `/usr/local/go/bin`, it should be `PATH=/usr/local/go/bin:${PATH}`.

Second, if you installing Go, you should do it the right way and setup all your PATHs, eg. `GOPATH`. Imo

CC/Ping @GodloveD, @devprincess

quick_start.rst
```diff
@@ -25,8 +25,10 @@ Quick Installation Steps
 
 You will need a Linux system to run Singularity.
 
-See the :ref:`installation page <installation>` for information about installing
-Singularity.
+.. note::
+     This is only a short manual for Quick Installation. For details on
+     different methods of Installation, versions, and RPM etc, please refer to
```
GodloveD: Change Installation to lowercase (installation) on line 30 plz

quick_start.rst
```diff
@@ -48,46 +50,52 @@ You must first install development libraries to your host. Assuming Ubuntu
         git
 
 .. note::
-    Note that ``squashfs-tools`` is only a dependency for commands that build 
-    images. The ``build`` command obviously relies on ``squashfs-tools``, but 
-    other commands may do so as well if they are run using container images 
+    Note that ``squashfs-tools`` is only a dependency for commands that builds
```
GodloveD: ~builds~ build (it was right the first time since _commands_ is plural.  😄 )

quick_start.rst
```diff
@@ -48,46 +50,52 @@ You must first install development libraries to your host. Assuming Ubuntu
...
+extract the archive to ``/usr/local`` (or use other instructions on go installation
 page).
 
 .. code-block:: none
 
-    $ export VERSION=1.12 OS=linux ARCH=amd64 && \
-        wget https://dl.google.com/go/go$VERSION.$OS-$ARCH.tar.gz && \
-        sudo tar -C /usr/local -xzvf go$VERSION.$OS-$ARCH.tar.gz && \
-        rm go$VERSION.$OS-$ARCH.tar.gz
+    $  sudo tar -C /usr/local -xzvf go1.12.5.linux-amd64.tar.gz
```
GodloveD: I'm not convinced that that is an improvement over the `wget` method.  With the previous version, I can simply copy and paste the 4 line command and I'm done.  With these changes I have to navigate away from this page, manually download a file with a click, figure out where it went, and then copy it to where I want it before I even execute the command above.  Even worse, this workflow is not captured in the instructions here so I have to just figure it out.

quick_start.rst
GodloveD: I'm OK with this more minimal method.  Maybe it is not "the right way", but maybe this _is_ the right way now that Go modules exist.  And it's not really our responsibility to make sure that people set their Go env up properly.  We just want to make sure they can install Singularity.  We also give several links to Go docs so people can look there if they want the official Golang way to install Go.

quick_start.rst
```diff
@@ -115,7 +125,7 @@ Singularity must be installed as root to function properly.
 Overview of the Singularity Interface
 -------------------------------------
 
-Singularity’s :ref:`command line interface <command-usage>` allows you to build
+Singularity’s command line interface allows you to build
```
GodloveD: Why remove this link?  Is it dead?  Can you just fix it instead?

quick_start.rst
```diff
@@ -278,14 +287,12 @@ containers of interest on the `Container Library <https://cloud.sylabs.io/librar
     	library://sylabsed/examples/alpine
     		Tags: latest
 
-You can use the :ref:`pull <pull-command>` and :ref:`build <build-command>`
+You can use the `pull <https://www.sylabs.io/guides/3.2/user-guide/cli/singularity_pull.html>`_
+and `build <https://www.sylabs.io/guides/3.2/user-guide/cli/singularity_build.html>`_
```
GodloveD: We really don't want to put these external links in here imho because they are much more prone to break. We just want to fix the internal links if they are broken. Otherwise these are going to break every time we update the docs to a new version for instance. Fixing the internal links if much more future-proof.

quick_start.rst
```diff
@@ -278,14 +287,12 @@ containers of interest on the `Container Library <https://cloud.sylabs.io/librar
...
 commands to download pre-built images from an external resource like the
 `Container Library <https://cloud.sylabs.io/library>`_ or
 `Docker Hub <https://hub.docker.com/>`_.
 
-When called on a native Singularity image like those provided on the Container
-Library, ``pull`` simply downloads the image file to your system.
```
GodloveD, line 287: I don't understand this deletion.

quick_start.rst
```diff
@@ -317,10 +324,9 @@ container like so:
 
 Unlike ``pull``, ``build`` will convert your image to the latest Singularity
 image format after downloading it.
-
 ``build`` is like a “Swiss Army knife” for container creation. In addition to
 downloading images, you can use ``build`` to create images from other images or
-from scratch using a :ref:`definition file <container-recipes>`. You can also
+from scratch using a :ref:`definition file <definitionfiles>`. You can also
```
GodloveD, line 335: yay!!

quick_start.rst
```diff
@@ -572,15 +576,15 @@ the context of a definition file, so you are advised to exercise care.
...
+build a SIF file using a Singularity definition file. This also makes it easy to
+add files, environment variables, and install custom software, and still start
+from your base of choice (e.g., the Container Library).
 
 A definition file has a header and a body. The header determines the base
-container to begin with, and the body is further divided into sections that do
-things like install software, setup the environment, and copy files into the
-container from the host system.
+container to begin with, and the body is further divided into sections that
+performs things like software installation, environment setup, and copying files
```
GodloveD: ~performs~ -> perform (because _sections_ is plural).

GodloveD (changes requested):
Thanks for these changes @sushma-4.  I agree with most of them, but there are others I think should be reverted.  I think the `wget` command should remain as a more efficient way to download singularity.  Also I really don't think we want to replace internal links with external links.  If we do so we will need to update all of the external links every time we update the docs versions or all of our links will be pointing to older versions of the docs.  Better to just fix the internal references if they are busted.

quick_start.rst
sushma-4: Right, the commands are definitely easier to work with, but for somebody who's not good with commands(students I held workshop for, even myself in the very beginning) it was easier to understand what was happening the UI way, but I forgot there are people who would simply want to copy the commands, so I added back the commands and wrote comments alongside on what each of those meant. Is that okay?

quick_start.rst
sushma-4: The link to Go Installation page is included in this section. Thank you @GodloveD

quick_start.rst
sushma-4: Agree! I totally wanted to do it with internal links.... because these are auto generated pages.. they don't have a ``rst`` page in the Repo. Is there any other way we can do it?

quick_start.rst
sushma-4: OMG I shouldn't mess with English Grammar

quick_start.rst
sushma-4: The link was probably from the page that was replaced by auto generated cli docs.. I manually checked every doc page, but couldn't find it.. maybe I should create a new link to the main CLI page and link it here.

quick_start.rst
sushma-4: Me neither :/ Maybe I was trying to add it after the command and forgot. Sorry

quick_start.rst
GodloveD: You need to make a reference to the place you want to link to and then use that reference in the RST docs.  There are lots of existing examples sprinkled throughout the docs right now.

jchavezlavalle: @sushma-4 you should just add a variable in replacements.py and then reference where needed, with that we would avoid having to update in every single place :+1:

sushma-4: I actually tried using ``replace`` so that a variable can be replaced, just tried the method you suggested, but while reading the URL, its not substituting the value.

Its getting read as special characters:
https://www.sylabs.io/guides/ **%7Bversion%7D** /

jchavezlavalle: yes this was solved by considering the following changes: on the `replacements.py` consider this declaration:

```
variable_replacements = {
    "{InstallationVersion}" : "3.2.0"
    "\{version\}" : "3.2"
}
``` 
and then on the rst just call the variable on the link ref as:

```
`pull <https://www.sylabs.io/guides/\{version\}/user-guide/cli/singularity_pull.html>`_
```

WestleyK (approved):
Looks good to me :+1:

quick_start.rst
GodloveD: Wait.  I don't think we want to do this either.  RST provides a shorthand way to make internal links right?  There is no need to spell out the whole external link like this.  Plugging in the version variable helps, but it's still not a great solution. 

Look at line 13 in the `quickstart.rst`
```
:ref:`requesting an installation help page <installation-request>`
```
This corresponds to a reference in `installation.rst` at line 497:
```
.. _installation-request:
```
This is the strategy that I think we need to use throughout to avoid broken links.  This way even if we change the location of the entire web page the internal links will still not break.

GodloveD (changes requested):
I still think the internal links should not be referencing an external URI.  See my comments inline for details.

installation.rst
```diff
@@ -257,8 +257,8 @@ Remove an old version
 When you run ``sudo make install``, the command lists files as they are
 installed. They must all be removed in order to completely remove Singularity.
 
-For example, in a standard installation of Singularity 3.0.1 and beyond (when 
-building from source) you must remove all of these files and directories to 
+For example, in a standard installation of Singularity 3.0.1 and beyond (when
+building from source) you must remove all of these files and directories to
 completely remove Singularity.
```
GwendolynK-zz, line 260:
```suggestion

```
This exact phrase was used in the sentence before. There isn't a need to repeat. One other note, to remain grammatically correct when removing this section of the sentence, we also need to remove the 'to' that proceeds it on the line right before. 
End sentence should be "For example, in a standard installation of Singularity 3.0.1 and beyond (when building from source) you must remove all of these files and directories."

quick_start.rst
```diff
@@ -25,8 +25,10 @@ Quick Installation Steps
 
 You will need a Linux system to run Singularity.
 
-See the :ref:`installation page <installation>` for information about installing
-Singularity.
+.. note::
+     This is only a short manual for Quick Installation. For details on
+     different methods of installation, versions, and RPM etc, please refer to
```
GwendolynK-zz, line 30:
```suggestion
     different methods of installation, versions, RPMs, etc, please refer to
```
@WestleyK is this still accurate technically? It is grammatically now...

quick_start.rst
```diff
@@ -48,46 +50,55 @@ You must first install development libraries to your host. Assuming Ubuntu
         git
 
 .. note::
-    Note that ``squashfs-tools`` is only a dependency for commands that build 
-    images. The ``build`` command obviously relies on ``squashfs-tools``, but 
-    other commands may do so as well if they are run using container images 
+    Note that ``squashfs-tools`` is only a dependency for commands that build
+    images. The ``build`` command obviously relies on ``squashfs-tools``, but
+    other commands may do so as well if they are run using container images
```
GwendolynK-zz:
```suggestion
    other commands may do so as well if they are ran using container images
```
or 
```suggestion
    other commands may do so as well if they are running using container images
```

quick_start.rst
```diff
@@ -330,12 +342,11 @@ Singularity.
 Interact with images
 --------------------
 
-You can interact with images in several ways. It is not actually necessary to
-``pull`` or ``build`` an image to interact with it. The commands listed here
-will work with image URIs in addition to accepting a local  path to an image.
+You can interact with images in several ways each of which can accept image URIs
```
GwendolynK-zz:
```suggestion
You can interact with images in several ways, each of which can accept image URIs
```
Helps sentence flow.

quick_start.rst
```diff
@@ -418,7 +431,7 @@ Running a container
 
 Singularity containers contain :ref:`runscripts <runscript>`. These are user
 defined scripts that define the actions a container should perform when someone
-runs it. The runscript can be triggered with the :ref:`run <run-command>`
+runs it. The runscript can be triggered with the `run <https://www.sylabs.io/guides/\{version\}/user-guide/cli/singularity_run.html>`_
 command, or simply by  calling the container as though it were an executable.
```
GwendolynK-zz:
```suggestion
 command, or simply by calling the container as though it were an executable.
```
extra space

quick_start.rst
```diff
@@ -572,15 +582,15 @@ the context of a definition file, so you are advised to exercise care.
...
+add files, environment variables, and install custom software, and still start
+from your base of choice (e.g., the Container Library).
 
 A definition file has a header and a body. The header determines the base
-container to begin with, and the body is further divided into sections that do
-things like install software, setup the environment, and copy files into the
-container from the host system.
+container to begin with, and the body is further divided into sections that
+perform things like software installation, environment setup, and copying files
+into the container from host system etc.
```
GwendolynK-zz:
```suggestion
into the container from host system, etc.
```

installation.rst
```diff
@@ -16,15 +16,15 @@ Overview
 
 Singularity runs on Linux natively and can also be run on Windows and Mac
 through virtual machines (VMs). Here we cover several different methods of
-installing Singularity (>={InstallationVersion}) on Linux and also give methods 
-for downloading and running VMs with singularity pre-installed from `Vagrant 
+installing Singularity (>={InstallationVersion}) on Linux and also give methods
+for downloading and running VMs with singularity pre-installed from `Vagrant
```
GwendolynK-zz:
```suggestion
for downloading and running VMs with Singularity pre-installed from `Vagrant
```

GwendolynK-zz (changes requested):

installation.rst
sushma-4: Thanks for the review @GwendolynK 
Does this sound better?
 
In a standard installation of Singularity 3.0.1 and beyond (when
building from source), the command ``sudo make install`` lists all the files as
they are installed. You must remove all of these files and directories to
completely remove Singularity.

quick_start.rst
sushma-4: I'm referring to https://sylabs.io/guides/3.2/user-guide/installation.html#build-and-install-an-rpm on our installation page. It should be just RPM IMO

GwendolynK-zz (approved):

PR #207, opened 2019-07-08
Updated labels section example -> 3.2

### LINKED ISSUE #205: Equal sign on labels section -> env & metadata section
## Which Document page:

Environment and metadata section

As seen here -> https://github.com/sylabs/singularity/issues/3862

jchavezlavalle:
This fixes or addresses the following GitHub issues:

- Fixes #205

### DIFF
--- a/definition_files.rst
+++ b/definition_files.rst
@@ -450,6 +450,10 @@ Consider the example from the def file above:
         Author d@sylabs.io
         Version v0.0.1
 
+Note that labels are defined by a key-value pair and are delimited by a space character.
+On the previous example, the first variable label name is ``Author``` with a
+value of ``d@sylabs.io``.
+
 The easiest way to see labels is to inspect the image:
 
 .. code-block:: none
--- a/definition_files.rst
+++ b/definition_files.rst
@@ -450,11 +450,20 @@ Consider the example from the def file above:
         Author d@sylabs.io
         Version v0.0.1
 
-Note that labels are defined by a key-value pair and are delimited by a space character.
-On the previous example, the first variable label name is ``Author``` with a
+Note that labels are defined by a key-value pairs and are delimited by the first space character.
+On the previous example, the first label name is ``Author``` with a
 value of ``d@sylabs.io``.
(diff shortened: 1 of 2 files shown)


ikaneshiro (approved):
lgtm!

definition_files.rst
```diff
@@ -450,6 +450,10 @@ Consider the example from the def file above:
         Author d@sylabs.io
         Version v0.0.1
 
+Note that labels are defined by a key-value pair and are delimited by a space character.
```
Flamefire: Shouldn't it be "labels are defined by a key-value pair**s**"?

To even more highlight how this works, could you add `MyLabel Hello World` and  "the **last** ~~variable~~ label name is ``MyLabel``` with a value of ``Hello World``."

This would nicely show that it is split on the FIRST space and spaces in the value are allowed. And removes the "variable" (I guess accidentally added)

jchavezlavalle: that's a good example, thanks, working on it :+1:

definition_files.rst
```diff
@@ -450,7 +450,20 @@ Consider the example from the def file above:
         Author d@sylabs.io
         Version v0.0.1
 
-The easiest way to see labels is to inspect the image:
+Note that labels are defined by a key-value pairs and are delimited by the first space character.
```
Flamefire: remove "a".

And I was first confused about the "delimited by the first space character.". Does this mean key-value? One label from the next? But can't think of a better wording

definition_files.rst
```diff
@@ -450,7 +450,20 @@ Consider the example from the def file above:
...
+On the previous example, the first label name is ``Author``` with a
+value of ``d@sylabs.io``.
+
+Consider the following example:
+
+.. code-block:: singularity
+
+    %labels
+        MyLabel Hello World
```
Flamefire, line 452: Add this to the example output below

Flamefire (changes requested):

Flamefire (approved):

Issue #184, opened 2019-05-30
Document the order which each section in a definition file is executed


jscook2345:
We should document all the possible sections in a definition file and the order that they are executed.

sushma-4:
Fixed in #208

PR #208, opened 2019-07-09
Bootstrap agents and Def file sections

### LINKED ISSUE #93: Document new bootstrap agents in section on definition files
## Which Document page:

On what page is the problem? 

https://www.sylabs.io/guides/3.0/user-guide/definition_files.html#header

### Expected results:

What are you expecting to see?

Documentation for bootstrap agents added in Singularity 3.0 - specifically `docker-daemon`, `docker-archive`, `oci` and `oci-archive`. 

### Actual results:

What is the problem, what's misspelled, or what's missing?

These are new bootstrap agents that are missing.

sushma-4:

Lists the sections of def file in the order they are executed , and more info newer Bootstrap agents introduced in v3.2

- Fixes #93,#184

### DIFF
--- a/appendix.rst
+++ b/appendix.rst
@@ -419,10 +419,10 @@ image and the new image during bootstrap.
 Overview
 """"""""
 
-A container from supporting OCI Registries - Eg: Azure Container Registry, 
-Local registry servers etc can be used as your “base” image and later customized. 
-This allows you to build multiple images from the same starting point. For example,
-you may want to build several containers with the same custom python installation, 
+Using, this module, a container from supporting OCI Registries - Eg: ACR (Azure Container 
+Registry), local container registries, etc can be used as your “base” image and later 
+customized. This allows you to build multiple images from the same starting point. For 
+example, you may want to build several containers with the same custom python installation, 
 the same custom compiler toolchain, or the same base MPI installation. Instead of 
 building these from scratch each time, you could make use of ``oras`` to pull an 
 appropriate base container and then build new containers by adding customizations in 
--- a/appendix.rst
+++ b/appendix.rst
@@ -217,7 +217,7 @@ You can see them listed alphabetically below with their respective functionality
 #. **SINGULARITY_WRITABLE_TMPFS**: Makes the file system accessible as read-write with non-persistent data (with overlay support only). Default is set to false.
 
 
-.. _build-modules:
+.. _buildmodules:
 
 Build Modules
 -------------
@@ -257,7 +257,7 @@ use.
 
     From: <entity>/<collection>/<container>:<tag>
 
-The From keyword is mandatory. It specifies the container to use as a base.
+The ``From`` keyword is mandatory. It specifies the container to use as a base.
 ``entity`` is optional and defaults to ``library``. ``collection`` is
 optional and defaults to ``default``. This is the correct namespace to use for
 some official containers (``alpine`` for example). ``tag`` is also optional and
@@ -313,7 +313,7 @@ use.
 
     From: <registry>/<namespace>/<container>:<tag>@<digest>
 
-The From keyword is mandatory. It specifies the container to use as a base.
+The ``From`` keyword is mandatory. It specifies the container to use as a base.
 ``registry`` is optional and defaults to ``index.docker.io``. ``namespace`` is
 optional and defaults to ``library``. This is the correct namespace to use for
 some official containers (ubuntu for example). ``tag`` is also optional and will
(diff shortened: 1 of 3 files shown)


appendix.rst
```diff
@@ -414,6 +410,41 @@ within the container called ``/.singularity.d/bootstrap_history``. Singularity
...
+
+
+``oras`` bootstrap agent
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Overview
+""""""""
+
+A container from supporting OCI Registries; Eg: Azure Container Registry, 
+Local Registry Servers etc can be used as your “base” and later customized.
```
jchavezlavalle: Maybe this part is not well worded?

appendix.rst
```diff
@@ -414,6 +410,41 @@ within the container called ``/.singularity.d/bootstrap_history``. Singularity
...
+
+The Bootstrap keyword is always mandatory. It describes the bootstrap module to
+use.
+
+.. code-block:: singularity
+
+    From: oras://registry/namespace/image:tag
+
+The ``From`` keyword is mandatory. It specifies the container to use as a base.
+Mentioning ``tag`` to refer to the version of image you want to use is mandatory.
```
jchavezlavalle: Maybe would be better to say: "The `tag` is also mandatory and it describes the image version you want to use"

appendix.rst
```diff
@@ -732,3 +764,60 @@ into the core operating system. It is a best practice to supply only the bare
...
+The ``From`` keyword is mandatory and applies to these modules in the same nature
+as described for other Bootstrap agents.
+
+.. _scratch-agent:
+
+``scratch`` bootstrap agent
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Through all the Bootstrap agents mentioned above, you were essentially building
+over a base(parent) image pulled from either Library/Docker/Shub/Oras etc, but
```
jchavezlavalle, line 799: I would remove the "etc"

appendix.rst
```diff
@@ -732,3 +764,60 @@ into the core operating system. It is a best practice to supply only the bare
...
+Singularity offers support to create even the base images or minimal images to
+create your custom containers.
+
+Overview
+""""""""
+
+This module allows you to take full control of the content inside your container,
```
jchavezlavalle, line 806: I'd start mentioning the name of the module, so "The bootstrap agent module allows you to take full control..."

definition_files.rst
```diff
@@ -498,13 +523,16 @@ After building the help can be displayed like so:
 Multi-Stage Builds
 ------------------
 
-Singularity 3.2 introduces multi-stage builds where one environment can be used for compilation, then the resulting binary can be copied into a final environment.  This allows a slimmer final image that does not require the entire development stack.
+Starting with Singularity v3.2 multi-stage builds are supported where one environment
```
jchavezlavalle, line 526: can you refer to the version from the replacements.py config variable? Also.. is this for version 3.2 or 3.3? if it's for 3.2 then update the PR to point to 3.2 branch, otherwise leave it to master (master is 3.3 docs now)

jchavezlavalle (changes requested):
Please double check if this belongs to 3.2 or 3.3 documentation @sushma-4 also, I left some comments :+1: thanks!

appendix.rst
sushma-4: This line follows right after the heading(which contains the name) of the bootstrap agent. Also, for consistency I kept it the same way it was written previously for all other bootstrap agents

appendix.rst
sushma-4: I had not mentioned all the rest of the names like Zypper, busybox etc using which you can form a base as well, hence the etc

appendix.rst
sushma-4: Umm.. the explanation follows in the paragraph, its about pulling images from Singularity supporting OCI registries can be used, can you please point me to which words unclear?

definition_files.rst
sushma-4: It doesn't have to come through replacements.py because its a static value of the version. It was introduced in v3.2 and so essentially it will remain the same for following versions as well.

sushma-4:
All the changes here are for v3.3 @devprincess. So this remains on the master I guess

appendix.rst
jchavezlavalle: Eg: Azure Container Registry, 
Local Registry Servers etc

jchavezlavalle (approved):

PR #210, opened 2019-07-09
Add comment about filesystem permissions and update example to reflect.

rherban:
This example better reflects real world usage and accomplishes the same as /foo


rherban:
Fixes sylabs/singularity#3721

GodloveD (approved):
lgtm :shipit:

PR #227, opened 2019-07-30
added a warning about to admins about using the capability command group

### LINKED ISSUE #226: Warn users/admins that caps must be treated with caution
## Which Document page:

https://sylabs.io/guides/3.3/user-guide/security_options.html#linux-capabilities

### Expected results:

A note telling users and admins that granting linux capabilities to users within containers almost guarantees they will be able to break out and get root on the system of they want to.  Maybe add a few examples on use cases where caps should be considered (limiting attack surface on cloud native architectures for example) and caps should probably not be considered (multi-tenant HPC).  

### Actual results:

Nothing to this effect exists.

GodloveD:

It's important to let admins know what they are getting into if they decide to use the `capabilities` command. Previously, this information was implied, but this change makes it explicit.

- Fixes #226

### DIFF
--- a/security_options.rst
+++ b/security_options.rst
@@ -15,19 +15,32 @@ the security scope and context when running Singularity containers.
 Linux Capabilities
 ------------------
 
+.. note::
+     It is extremely important to recognize that **granting users Linux 
+     capabilities with the** ``capability`` **command group is usually identical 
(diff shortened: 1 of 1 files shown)


sushma-4 (approved):

rherban (approved):
LGTM

Issue #211, opened 2019-07-12
document need for cryptsetup-bin


GodloveD:
Looks like `cryptsetup-bin` has been added as a new dependency (presumably to support encrypted containers).  This will need to be documented in the 3.4 version of the docs.

WestleyK:
Please use `cryptsetup`, (instead of `cryptsetup-bin`).

`cryptsetup-bin` looks to be a dependency for `cryptsetup`, and they both need to be installed to function properly.

kmanalo:
I noticed this as well, should this message be altered for the configure then?

```
 checking: cryptsetup... no
unable to find the cryptsetup program, is the package cryptsetup-bin installed?
```

dtrudg:
> Please use `cryptsetup`, (instead of `cryptsetup-bin`).
> 
> `cryptsetup-bin` looks to be a dependency for `cryptsetup`, and they both need to be installed to function properly.

@WestleyK - I'm not sure this is the case. `cryptsetup` on debian based distros includes recommends on initramfs related packages that we don't need. The `cryptsetup-bin` command line package should be sufficient, and avoid bringing in some things that are only needed for LUKS encryption on the host.

Can you elaborate on what error you see (if any) with `cryptsetup-bin` and without `cryptsetup`?

dtrudg:
Closing this. I'm pretty confident `cryptsetup-bin` is sufficient here. Please reopen with detail if issues found anywhere.

PR #234, opened 2019-09-04
Add cryptsetup-bin build dependency for Singularity 3.4

hans:
I just encountered a missing package when following the Singularity build instructions with the new 3.4 release. Installing `cryptsetup-bin` fixes things.

### DIFF
--- a/installation.rst
+++ b/installation.rst
@@ -58,7 +58,8 @@ similar with other package managers.
         libseccomp-dev \
         wget \
         pkg-config \
-        git
+        git \
+        cryptsetup-bin
 
 ``yum``
 


installation.rst
```diff
@@ -58,7 +58,8 @@ similar with other package managers.
         libseccomp-dev \
         wget \
         pkg-config \
-        git
+        git \
+        cryptsetup-bin
```
WestleyK, line 62: I suggest using `cryptsetup`, because `cryptsetup-bin` looks to be a dependency for `cryptsetup`

```suggestion
        cryptsetup
```

WestleyK (commented):
Thanks for adding this, I just suggest using `cryptsetup`, otherwise looks good. Also looks like this closes https://github.com/sylabs/singularity-userdocs/issues/211.

PR #241, opened 2019-09-18
Removed a duplicate entry in index.rst

alalazo:
The "Container Security <security>" entry was duplicated in `index.rst` and shown twice in the sidebar.


WestleyK (approved):
Looks good to me! Good catch!

PR #245, opened 2019-09-27
cryptsetup needed for installation on Ubuntu 18.04

lpryszcz:

Write your description of the PR here. Be sure to include as much background,
and details necessary for the reviewers to understand exactly what this is
fixing or enhancing.

- Fixes #

### DIFF
--- a/quick_start.rst
+++ b/quick_start.rst
@@ -47,7 +47,8 @@ You must first install development libraries to your host. Assuming Ubuntu
         libseccomp-dev \
         wget \
         pkg-config \
-        git
+        git \
+        cryptsetup
 
 .. note::
     Note that ``squashfs-tools`` is only a dependency for commands that build


jmstover:
`cryptsetup` isn't a build requirement now. It's still looked for, but the build doesn't fail if it isn't found.

I'll leave this as I don't think it's a *bad* idea to have it at build time... I'll let others decide on it's inclusion at this point in the quick start though.

@jscook2345 @rherban 

*Edit:* Maybe ... wasn't this done???

jmstover (approved):
This looks fine to me. I'll let others decide on whether to merge it or not as `cryptsetup` not existing at this point in `master` isn't a build failure.

WestleyK (approved):
This looks good. Maybe at one point (not necessary in this PR) we should add a note that `cryptsetup` is only required for building encrypted containers, just like theres a note for `squashfs`.

CC @ikaneshiro

PR #246, opened 2019-10-01
Correcting one incorrect package name and one typo

justbennet:
Instructions were asking for libgpgme11-dev as dependency; apt-get substitutes libgpgme-dev; correcting here, at source.

Correctly spelled Singularity.


dtrudg (approved):
LGTM

PR #251, opened 2019-10-02
Bump 2.5 install VER to 2.5.2

### LINKED ISSUE #135: missing step in 2.5 install guide
Please fix the following issue:

https://github.com/sylabs/singularity/issues/1763

It is a bit of an old issue, but since we have the docs out there still I think it would be good to update it!

dtrudg:

Make the 2.5 docs refer to the latest/last 2.5.2 release in that series

- Fixes #135 

(See comment on #135 why autogen.sh is not required)

### DIFF
--- a/installation.rst
+++ b/installation.rst
@@ -92,7 +92,7 @@ releases <https://github.com/sylabs/singularity/releases>`_ page to ``/usr/local
 
 .. code-block:: none
 
-    $ VER=2.5.1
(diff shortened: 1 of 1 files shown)


jmstover (approved):
LGTM

dtrudg:
Note - this will fail CI as we only added the CI config for `master` and `3.4` to avoid work to remove lint on older branches.

PR #300, opened 2020-01-08
Remove detailed installation section (to admindocs)

### LINKED ISSUE #289: Consider moving installation instructions to admindocs only?
The installation instructions in the userdocs repo here, and the admindocs repo (which is undergoing renovation for 3.5) are going to be duplicating content.

I'm thinking it might be better to move the majority of installation content to the admin docs, and just leave a reference and maybe the minimal install from source instructions here?

Any thoughts @ikaneshiro ?

dtrudg:

Remove detailed installation section to the admindocs, so that we don't have so much duplicate information, and it's co-located with the configuration documentation that system administrators will be interested in.

- Fixes #289


quick_start.rst
```diff
@@ -657,3 +660,67 @@ do with Singularity!
...
+        - Security: A discussion of security concerns is discussed at
+        https://www.sylabs.io/guides/{adminversion}/admin-guide/admin_quickstart.html
+
+        - Installation:
+        https://www.sylabs.io/guides/{adminversion}/admin-guide/installation.html
+
+    If you have questions about any of the above, you can email the open source
+    list (singularity@lbl.gov), join the open source slack channel
+    (singularity-container.slack.com), or contact the organization that supports
+    Singularity directly to get a human response (sylabs.io/contact). I can do
```
ikaneshiro: I feel like this implies that the other contact methods _won't_ get a human response
```suggestion
    Singularity directly (sylabs.io/contact). I can do
```

dtrudg: Yeah - this was just moved, not edited here - but I think this is fair.

ikaneshiro (approved):

PR #306, opened 2020-01-10
Correct in memory encryption statement

dtrudg:

In memory part of encryption statement in the 3.4 docs is potentially misleading, so remove it. It was intended to indicate the container isn't _decrypted entirely into RAM before being run_. However, caching etc. and the need to run decrypted code means there is content in RAM - so the wording is misleading.


ikaneshiro (approved):
CI failure looks to just be a CI issue

dtrudg:
Going to merge against the CI :x: as that's on the markdown, and there is no change there in this PR>

PR #336, opened 2020-07-19
Correct overlay image upper *and* work dir creation.

dtrudg:

As discussed in https://github.com/hpcng/singularity/issues/5430
we need to create both upper and work dirs in an overlay image
given the privilege / capability set the RPC server now has.


cclerget (approved):
LGTM

PR #384, opened 2021-02-28
Remove inaccurate claim about operating systems

mkenigs:
Operating system includes kernel, and singularity does not allow choosing a different kernel


brianclemens:
"Distribution userland" might be more accurate, but pedantic and more difficult for the folks that message is intended to go to.

PR #411, opened 2021-10-07
Document restrictions/workaround regarding build and bind option

### LINKED ISSUE #410: Document restrictions/workaround regarding build and bind option
There is no mention regarding the requirement of having the mount destinations to exist when using `--bind` during build. Also put a workaround note.

cclerget:

- Fixes #410
- Related to https://github.com/hpcng/singularity/issues/6181

### DIFF
--- a/build_a_container.rst
+++ b/build_a_container.rst
@@ -296,11 +296,23 @@ The ``--update`` option is only valid when used with sandbox containers.
 This flag allows you to mount the Nvidia CUDA libraries of your host into your build
 environment. Libraries are mounted during the execution of ``post`` and ``test`` sections.
 
+.. note::
+
+    This option can't be set via the environment variable `SINGULARITY_NV`.
+    Singularity will attempt to bind binaries listed in SINGULARITY_CONFDIR/nvliblist.conf,
+    if the mount destination doesn't exist inside the container, they are ignored.
+
 ``--rocm``
 ==========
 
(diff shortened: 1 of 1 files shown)


build_a_container.rst
```diff
@@ -296,12 +296,20 @@ The ``--update`` option is only valid when used with sandbox containers.
 This flag allows you to mount the Nvidia CUDA libraries of your host into your build
 environment. Libraries are mounted during the execution of ``post`` and ``test`` sections.
 
+.. warning::
```
DrDaveD: I think that a Note is sufficient for this, not a warning.

build_a_container.rst
```diff
@@ -296,12 +296,20 @@ The ``--update`` option is only valid when used with sandbox containers.
 This flag allows you to mount the Nvidia CUDA libraries of your host into your build
 environment. Libraries are mounted during the execution of ``post`` and ``test`` sections.
```
DrDaveD, line 297: According to my testing of the --nv option (and presumably the --rocm option is similar) silently ignores doing mounts of the /usr/bin files if a mount point is not included in the container.  I think that should be documented, probably along with a pointer to the /etc/singularity file that contains the list of files.

DrDaveD (changes requested):

DrDaveD (approved):

Issue #59, opened 2019-01-09
Pycache folder should be git ignored


jchavezlavalle:
Related to PR #54 

## Which Document page:

In general, related to the whole project structure.

### Expected results:

`__pycache__` folder should not be listed as a part of the changes.

### Actual results:

After running `make html` command (to generate html content from rst) the python script from pygments generates a  `__pycache__` folder that should not be listed as a part of the changes.

PR #429, opened 2021-12-09
Add new SINGULARITY_ env vars, from sylabs 59

edytuk:
It takes a little bit from sylabs PR

- sylabs/singularity-userdocs#59

### DIFF
--- a/appendix.rst
+++ b/appendix.rst
@@ -70,7 +70,20 @@ You can see them listed alphabetically below with their respective functionality
 
 #. **SINGULARITY_DOCKER_PASSWORD**: To specify the password for docker authentication.
 
-#. **SINGULARITY_DROP_CAPS**: To specify a list (comma separated string) of capabilities to be dropped. Default is an empty string.
+#. **SINGULARITY_DOWNLOAD_CONCURRENCY**: To specify how many concurrent streams
+   when downloading (pulling) an image from cloud library.
+
+#. **SINGULARITY_DOWNLOAD_PART_SIZE**: To specify the size of each part (bytes)
+   when concurrent downloads are enabled.
+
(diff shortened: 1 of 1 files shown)


appendix.rst
```diff
@@ -70,7 +70,17 @@ You can see them listed alphabetically below with their respective functionality
 
 #. **SINGULARITY_DOCKER_PASSWORD**: To specify the password for docker authentication.
 
-#. **SINGULARITY_DROP_CAPS**: To specify a list (comma separated string) of capabilities to be dropped. Default is an empty string.
+#. **SINGULARITY_DOWNLOAD_CONCURRENCY**: To specify how many concurrent streams
```
DrDaveD, line 73: These 3 will need to be removed when the library interface is removed, but for now they match what is in the apptainer master branch.

appendix.rst
```diff
@@ -147,7 +164,11 @@ You can see them listed alphabetically below with their respective functionality
 
 #. **SINGULARITY_NO_PRIVS**: To drop all the privileges from root user in the container. Default is set to false.
 
-#. **SINGULARITY_NV**: To enable experimental Nvidia support. Default is set to false.
+#. **SINGULARITY_NV**: To enable Nvidia GPU support. Default is
+   set to false.
+
+#. **SINGULARITY_NVCCLI**: To use nvidia-container-cli for container GPU setup
+   (experimental).
```
cclerget: For apptainer, this option is only supported for unprivileged run (user namespace) and root user, a note mentioning that would avoid some confusion

DrDaveD:
```suggestion
   (experimental, only unprivileged).
```

DrDaveD:
Since this is 4 commits, when we merge we should use "squash and merge"

cclerget (approved):

PR #44, opened 2018-12-17
Added Contributing to 3.0 branch

WestleyK:
~~Not much changed from 2.X. mostly copy-paste~~ :smile:

Not much copy-paste. more changes! :smile:

Some changes include:
  - `Contributing to Documentation` is above `Contributing to Singulairty`.
  - Less Github tutorial, more readable.
  - Issue reporting.

<br>

This PR will fix/close:

  - Fixes #36

Issue #32, opened 2018-11-26
Cgroups documentation


hearnsj:
Version of Singularity:
3.0.1-1.fc29

Expected behavior

Documentation. At page https://syslabs.io/guides/3.0/user-guide/cgroups.html the example TOML file is said to be at location /usr/local/etc/singularity/cgroups/cgroups.toml

Actual behavior
On a Fedora 29 system the file is /etc/singularity/cgroups/cgroups.tomp

GodloveD:
I think this depends on how you do the installation.  Did you use the RPM method?  Want to suggest an edit in a PR? @hearnsj

GodloveD:
Heya @hearnsj I just went ahead and took the liberty of suggesting a fix in #181.  Thanks for the suggestion!

PR #431, opened 2021-12-10
Tidy up endpoint.rst, from sylabs 44

edytuk:
This pulls in sylabs PR

- sylabs/singularity-userdocs#44
 which fixed
- sylabs/singularity-userdocs#32


DrDaveD (approved):
This whole section will need a big rewrite but I don't think this makes it any harder to do that.

